PR #11, opened 2024-04-05
add some tests functions

yucongalicechen:
- added test_get_grid_points.py, test_set_distances_at_angle.py
- Slight mistake in functions.py: when we select the exit point it should be above the grid point, but when the grid point is at the bottom (0, -radius), we should compare the two roots instead of comparing with the grid point


diffpy/labpdfproc/tests/test_set_distances_at_angle.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+
+from diffpy.labpdfproc.functions import Gridded_circle
+
+
+def test_set_distances_at_angle():
+    actual_gs = Gridded_circle(radius=1, n_points_on_diameter=3, mu=1)
+
+    # tests for acute angles
```
sbillinge, line 9: when you do multiple tests it is better to use the `@pytest.mark.paramatrize()` decorator on the test function and run different examples as a list of tuples.  You may be able to google how to do this.

sbillinge (commented):
good start, please see my inline comments.

PR #13, opened 2024-04-06
update functions.py

yucongalicechen:
- Update functions.py: modify the algorithm to select the correct grid point by comparing the two y-coordinates of points on the circle instead of comparing them with the y-coordinate of the grid point.

### DIFF
--- a/src/diffpy/labpdfproc/tests/test_functions.py
+++ b/src/diffpy/labpdfproc/tests/test_functions.py
@@ -1,0 +1,24 @@
+import numpy as np
+import pytest
+
+from diffpy.labpdfproc.functions import Gridded_circle
+
+
+def test_get_grid_points():
+    actual_gs = Gridded_circle(radius=1, n_points_on_diameter=3, mu=1)
+    expected_gs = Gridded_circle()
+    expected_gs.grid = {(0.0, -1.0), (0.0, 0.0), (-1.0, 0.0), (1.0, 0.0), (0.0, 1.0)}
+    expected_gs.total_points_in_grid = 5
+    assert actual_gs.grid == expected_gs.grid
+    assert actual_gs.total_points_in_grid == expected_gs.total_points_in_grid
+
+
+@pytest.mark.parametrize("angle, expected_distances", [
+    (25, [0, 2, 1.81261557406, 2, 0.845236523481399]),
+    (90, [0, 2, 0, 2, 2]),
+    (140, [0, 2, 0, 3.5320888862379562, 1.2855752193730785]),
+])
+def test_set_distances_at_angle(angle, expected_distances):
+    actual_gs = Gridded_circle(radius=1, n_points_on_diameter=3, mu=1)
+    actual_gs.set_distances_at_angle(angle)
+    assert actual_gs.distances == pytest.approx(expected_distances, rel=1e-4, abs=1e-6)
--- a/src/diffpy/labpdfproc/tests/test_functions.py
+++ b/src/diffpy/labpdfproc/tests/test_functions.py
@@ -3,22 +3,63 @@ import pytest
 
 from diffpy.labpdfproc.functions import Gridded_circle
 
+params1 = [
+    [[1, 3, 1, {(0.0, -1.0), (0.0, 0.0), (-1.0, 0.0), (1.0, 0.0), (0.0, 1.0)}, 5]],
+    [[0.2, 8, 1, {(-0.08571428571428572, -0.14285714285714285), (-0.02857142857142858, -0.14285714285714285), (0.02857142857142858, -0.14285714285714285), (0.08571428571428574, -0.14285714285714285), (-0.14285714285714285, -0.08571428571428572), (-0.08571428571428572, -0.08571428571428572), (-0.02857142857142858, -0.08571428571428572), (0.02857142857142858, -0.08571428571428572), (0.08571428571428574, -0.08571428571428572), (0.14285714285714285, -0.08571428571428572), (-0.14285714285714285, -0.02857142857142858), (-0.08571428571428572, -0.02857142857142858), (-0.02857142857142858, -0.02857142857142858), (0.02857142857142858, -0.02857142857142858), (0.08571428571428574, -0.02857142857142858), (0.14285714285714285, -0.02857142857142858), (-0.14285714285714285, 0.02857142857142858), (-0.08571428571428572, 0.02857142857142858), (-0.02857142857142858, 0.02857142857142858), (0.02857142857142858, 0.02857142857142858), (0.08571428571428574, 0.02857142857142858), (0.14285714285714285, 0.02857142857142858), (-0.14285714285714285, 0.08571428571428574), (-0.08571428571428572, 0.08571428571428574), (-0.02857142857142858, 0.08571428571428574), (0.02857142857142858, 0.08571428571428574), (0.08571428571428574, 0.08571428571428574), (0.14285714285714285, 0.08571428571428574), (-0.08571428571428572, 0.14285714285714285), (-0.02857142857142858, 0.14285714285714285), (0.02857142857142858, 0.14285714285714285), (0.08571428571428574, 0.14285714285714285)}, 32]]
+]
 
-def test_get_grid_points():
-    actual_gs = Gridded_circle(radius=1, n_points_on_diameter=3, mu=1)
-    expected_gs = Gridded_circle()
-    expected_gs.grid = {(0.0, -1.0), (0.0, 0.0), (-1.0, 0.0), (1.0, 0.0), (0.0, 1.0)}
-    expected_gs.total_points_in_grid = 5
-    assert actual_gs.grid == expected_gs.grid
-    assert actual_gs.total_points_in_grid == expected_gs.total_points_in_grid
+@pytest.mark.parametrize('params1', params1)
+def test_get_grid_points(params1):
+    for param_set in params1:
+        radius, n_points_on_diameter, mu, expected_grid, expected_total_points = param_set
+        actual_gs = Gridded_circle(radius=radius, n_points_on_diameter=n_points_on_diameter, mu=mu)
+        assert actual_gs.grid == expected_grid
+        assert actual_gs.total_points_in_grid == expected_total_points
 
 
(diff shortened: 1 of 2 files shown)


yucongalicechen:
- Updated _get_grid_points in functions.py where we prefer to have the grid as a set instead of an array.
- Included test_functions.py with tests functions for get_grid_points and set_distances_at_angle.

sbillinge (commented):
this looks great, nice job.

- I wonder if we should check intermediate points in the gridding, so go from 3 to 5 on the grid as right now we are only testing points at the center and on the circumference.
- Also for the angle, do we want to test points coming from different quadrants to be sure about the results?
- Please also see my inline comment

The PR is failing CI.  Did you sync your main and then merge main into this branch?  If you did, then there is something else wrong.  You can see by clicking on "details" and see where it failed.

sbillinge:
I see that you are failing a bunch of pre-commit tests too.  This should not happen if the pre-commit is passing on your local.  Please run `pre-commit install` in the top level directory to get this working.

yucongalicechen:
- For test_get_grid_points: checked grid_size = 3 and 8 (so both odd and even numbers - cases where grid points fall on the circle and fall inside circle).
- For test_set_distances_at_angle: checked distances for acute/right/obtuse angles for grid_size = 3 and 6 and radius = 1 and 0.3. I was going to include some extreme cases like angle = 89.9, but I thought that we're probably not going to calculate distances for these angles, so would this be necessary? Also, although the values of the distances are correct, the order of the distances is a bit weird, I will check about this.
- For test_set_muls_at_angle: values are correct. This probably won't be an issue when we use the function because we will always call set_distances function first, but it seems that I need to add the set_distances function before using set_muls, where there is a set_distances function inside set_muls (i.e., not sure if set_distances inside set_muls is working probably).
- There are two other functions get_path_length and get_entry_exit_coordinates in Class Gridded_circle. They are included in set_distances_at_angle, so I didn't write other tests for them.

src/diffpy/labpdfproc/tests/test_functions.py
```diff
@@ -0,0 +1,65 @@
...
...8572, -0.08571428571428572), (-0.02857142857142858, -0.08571428571428572), (0.02857142857142858, -0.08571428571428572), (0.08571428571428574, -0.08571428571428572), (0.14285714285714285, -0.08571428571428572), (-0.14285714285714285, -0.02857142857142858), (-0.08571428571428572, -0.02857142857142858), (-0.02857142857142858, -0.02857142857142858), (0.02857142857142858, -0.02857142857142858), (0.08571428571428574, -0.02857142857142858), (0.14285714285714285, -0.02857142857142858), (-0.14285714285714285, 0.02857142857142858), (-0.08571428571428572, 0.02857142857142858), (-0.02857142857142858, 0.02857142857142858), (0.02857142857142858, 0.02857142857142858), (0.08571428571428574, 0.02857142857142858), (0.14285714285714285, 0.02857142857142858), (-0.14285714285714285, 0.08571428571428574), (-0.08571428571428572, 0.08571428571428574), (-0.02857142857142858, 0.08571428571428574), (0.02857142857142858, 0.08571428571428574), (0.08571428571428574, 0.08571428571428574), (0.14285714285714285, 0.08571428571428574), (-0.08571428571428572, 0.14285714285714285), (-0.02857142857142858, 0.14285714285714285), (0.02857142857142858, 0.14285714285714285), (0.08571428571428574, 0.14285714285714285)}, 32]]
```
sbillinge, line 8: this set of values seems to be very specific.  How did you generate them and how do you know they are correct?  A better test is one where you know the right answer and generate  without using the function. For an odd number of points use 0.5.  

The even number of points specified is harder to test, so do it with a very small number of points, say 2 on each axis and then the points shold presumably be something like 1/3, 2/3.

src/diffpy/labpdfproc/tests/test_functions.py
```diff
@@ -0,0 +1,65 @@
+import numpy as np
+import pytest
+
+from diffpy.labpdfproc.functions import Gridded_circle
+
+params1 = [
+    [[1, 3, 1, {(0.0, -1.0), (0.0, 0.0), (-1.0, 0.0), (1.0, 0.0), (0.0, 1.0)}, 5]],
```
sbillinge, line 7: these are easier to read if they are a list of tuples, one for each test, then this tuple has two lists which are for the inputs and outputs.  So in this case it would be 
```
params1 = [
    ( [1, 3, 1], [{(...), (...), (...)}, 9] ),
    (...)
]
@pytest.mark.paramaterize("inputs, expected", params1)
def test_...():
    circle = Griddedcircle(inputs[0], inputs[1])
```
and so on.  This is much easier to read.

I may not have the syntax from `pytest.mark.paramatrize()` quite right so please look it up, but we would like the intent of hte tests to be easier to read in all your tests.

src/diffpy/labpdfproc/tests/test_functions.py
```diff
@@ -0,0 +1,65 @@
...
+]
+
+@pytest.mark.parametrize('params1', params1)
+def test_get_grid_points(params1):
+    for param_set in params1:
```
sbillinge, line 13: this is not how paramatrize works, you don't need this.

src/diffpy/labpdfproc/tests/test_functions.py
```diff
@@ -0,0 +1,65 @@
...
+        radius, n_points_on_diameter, mu, expected_grid, expected_total_points = param_set
+        actual_gs = Gridded_circle(radius=radius, n_points_on_diameter=n_points_on_diameter, mu=mu)
+        assert actual_gs.grid == expected_grid
+        assert actual_gs.total_points_in_grid == expected_total_points
+
+
+
+# values of distances are correct but im a bit confused about the order of distances
```
sbillinge, line 21: Decide in your test what behavior you want.  Do you care about order?  If not, you could use a set, or stick with a list but sort it by value before doing the assert, something like that.  If you care about order in the function then test this.

src/diffpy/labpdfproc/tests/test_functions.py
```diff
@@ -0,0 +1,65 @@
...
+params2 = [
+    [[1, 3, 1, [25, 90, 140],
```
sbillinge, line 23: you need to run `black` on everything to get the layout right.  But when you install `pre-commit` it will do it for you when you commit.  You will have to run the commit multiple times as when it runs black the commit fails, so just rerun it (if flake8 fails you have to fix the errors manually then rerun the commit).

src/diffpy/labpdfproc/tests/test_functions.py
```diff
@@ -0,0 +1,65 @@
...
+     [[0, 2, 1.81261557406, 2, 0.845236523481399],
+      [0, 2, 0, 2, 2],
+      [0, 2, 0, 3.5320888862379562, 1.2855752193730785]]]],
```
sbillinge, line 26: many of these numbers seem magical and too high in precision (is it important for your function that it gets a number correct to 11 significant figures?).  Can you manually compute a few special cases and use those as the test?  It may be harder here than for the grid points, so just find a reasonable balance, but don't use the function to compute these.

src/diffpy/labpdfproc/tests/test_functions.py
```diff
@@ -0,0 +1,65 @@
...
+    [[0.3, 6, 1, [14, 90, 120],
+      [[0.5480340818997349, 0.5354482281467614, 0.601285995493114, 0.6027483291753729, 0.6034715865980909, 0.578005471973011, 0.4947788959442535, 0.5960138080070901, 0.5562896419944887, 0.5686908551742735, 0.45192085857776143, 0.427513512087355, 0.5844112402440541, 0.47163573802449577, 0.5181699073501158, 0.3969934749877318],
+       [0.48, 0.6539387691339813, 0.7078775382679627, 0.41393876913398137, 0.5878775382679627, 0.5878775382679627, 0.84, 0.7739387691339814, 0.29393876913398137, 0.46787753826796274, 0.41393876913398137, 0.29393876913398137, 0.6539387691339813, 0.48, 0.7739387691339814, 0.12],
+       [0.2977588782576409, 0.5701702621142971, 0.7350953659368239, 0.31137200214817157, 0.5444870368999288, 0.6205639884457963, 0.9585605421126093, 0.8882109144555469, 0.24428786600141425, 0.45117231748269127, 0.4384011167518993, 0.2916941133177683, 0.747448953694039, 0.5259897328952432, 0.7834632586801662, 0.10679139675021132]]]
+    ]
+]
+
+@pytest.mark.parametrize('params2', params2)
+def test_set_distances_at_angle(params2):
+    for param_set in params2:
```
sbillinge, line 36: remove this for loop.

src/diffpy/labpdfproc/tests/test_functions.py
```diff
@@ -0,0 +1,65 @@
...
+     [[1, 0.01831563888, 0.02664293815, 0.01831563888, 0.18443225823],
+      [1, 0.01831563888, 1, 0.01831563888, 0.01831563888],
+      [1, 0.01831563888, 1, 0.00085519779, 0.07644754659]]]
+]
+@pytest.mark.parametrize('params3', params3)
+def test_set_muls_at_angle(params3):
+    radius, n_points_on_diameter, mu, angles, expected_muls = params3
+    actual_gs = Gridded_circle(radius=radius, n_points_on_diameter=n_points_on_diameter, mu=mu)
+    for angle, expected_mul in zip(angles, expected_muls):
+        # here it seems like I need to set distances first (can't assign distances only using set muls)
```
sbillinge, line 56: remember, figure out what you want your function to do then write the test to do it.  What does this function do and how is it used, and so what behavior do you want in different circumstances, then write the test to check that.  Don't be distracted by what the function actually does at the moment.  Let's just write the tests first, then we can rewrite any functions to make them have that behavior.

sbillinge (commented):
please see inline comments.

ideally we would have one test per branch and it would be easier to merge  these.  Just decide if you want to close this PR and do that, or just continue with this PR.   

Also, quick question, did you manage to merge upstream main?

sbillinge:
I htink this PR can be closed as it is replaced with other PRs.  If you agree, please can you close it?

PR #17, opened 2024-04-10
Tests for setting distances for each grid point

yucongalicechen:
Tests passed assertion. Checked distances for both odd and even grid points (i.e., points on the circle, in the center, and at each quadrant) and for acute, right, and obtuse angles.


sbillinge:
super!

PR #21, opened 2024-04-27
initial commit of issue 7: loading all input parameters into metadata

### LINKED ISSUE #27: load all the input parameters into metadata

yucongalicechen:
Closes #27 

Initial commit of loading all parameters into metadata, not sure if we need to load input_file, output_directory, force_overwrite, etc., since they seem to be less related with the function of insert_scattering_quantity?

### DIFF
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -36,7 +36,7 @@ def get_args():
     p.add_argument(
         "-x",
         "--xtype",
-        help=f"the quantity on the independnt variable axis. allowed "
+        help=f"the quantity on the independent variable axis. allowed "
         f"values: {*XQUANTITIES, }. If not specified then two-theta "
         f"is assumed for the independent variable. Only implemented for "
         f"tth currently",
@@ -54,7 +54,7 @@ def get_args():
         "-f",
         "--force-overwrite",
         action="store_true",
-        help="outputs will not overwrite existing file unless --force is spacified",
+        help="outputs will not overwrite existing file unless --force is specified",
     )
     args = p.parse_args()
     return args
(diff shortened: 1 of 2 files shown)


sbillinge:
> Initial commit of loading all parameters into metadata, not sure if we need to load input_file, output_directory, force_overwrite, etc., since they seem to be less related with the function of insert_scattering_quantity?

That is a great question.  I think the rule of thumb is whether this information may be useful to someone later.  

It is quite unlikely that knowing some location on someone's local hard-drive of where files were written might be useful, but not completely crazy.  For example, if it was in a group computer somewhere and you sent me a file by email, then you graduated and left and I wanted to point a new student to where the data are for example, the information would be in the file you had sent.  Since the cost/difficulty is low to add it to the file, I think we should do it.  However, it would clearly be much more useful if it was a full file-path and not just a relative path, so "/usr/yucong/home/analysis/till_project" is much more useful than ".", for example.

for the --force, and output_correciton then are of course not valuable and we don't need to save them.   In the same spirit, we should maybe save the input folder address as well as the output.   

To keep things lower case maybe we should be using `mu_d` instead of `muD`.

Thanks for catching the typo's.

sbillinge:
Thinking about it more, another design principle could be to make the code easier to maintain.  If we apply this design principle, it we would probably just write code that dumps all the arguments in some way.  This way, if we add arguments in the future they will automatically be captured without us having to remember to write the code that passes them into the metadata dict.   Perhaps that would be better, what do you think?  Of course, something in between where we can modify an arg (for example from relative to absolute file-path) before that happens?

sbillinge (commented):
OK, I thought about this more.  So we have better test coverage, let's make a `load_metadata()` function for this part and a `set_wavelength()` function.  Please then write tests for these functions.

load_metadata could just contain `metadata = vars(args)` as its functional step.

the set_wavelength is a different issue than this one, but while I think about it, if we are using the above for loading hte metadata, we need to load the correct wavelength into the `args.wavelength` variable even if this was not specified on the command line, and that i swhat the `set_wavelength` function will do. So for now just do the metadata loading this way here, but we will have a series of other functions that will update args.<something> parameters appropriately before we load the metadata.  Does that make sense?

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -36,7 +37,7 @@ def get_args():
     p.add_argument(
         "-x",
         "--xtype",
-        help=f"the quantity on the independnt variable axis. allowed "
+        help=f"the quantity on the independent variable axis. allowed "
         f"values: {*XQUANTITIES, }. If not specified then two-theta "
         f"is assumed for the independent variable. Only implemented for "
```
sbillinge, line 42: please check, do we have issues for implementing for other x-axes? if not, please could you make that issue?

sbillinge (commented):
this seems to be missing tests and news?

src/diffpy/labpdfproc/labpdfprocapp.py
yucongalicechen: just added this

PR #34, opened 2024-04-29
closes #22: specify an output directory name

### LINKED ISSUE #22: specify an output directory name
specify an output directory name. Expected behavior: code checks the dir is there, creates it if it isn't then and writes corrected files in there. Files exists/force behavior updated for this eventuality. Default could be "_corrected" or sthg.

yucongalicechen:
(Edited) Intermediate commit on specifying an output directory name.

Code's doing the following:
1. Specify an output directory name, default is cwd
2. Check the dir is there, creates it if it isn't
3. Write corrected files in there

output_dir is in absolute path and is loaded to args.output_directory.
Help message is edited but needs proofread (see comment below)

### DIFF
--- a/src/diffpy/labpdfproc/tests/test_tools.py
+++ b/src/diffpy/labpdfproc/tests/test_tools.py
@@ -9,7 +9,6 @@ params1 = [
     ([None], [Path.cwd().resolve()]),
     (["."], [Path.cwd().resolve()]),
     (["new_dir"], [Path.cwd().resolve() / "new_dir"]),
-    (["new_dir.py"], [Path.cwd().resolve() / "new_dir.py"]),
     (["existing_dir"], [Path.cwd().resolve() / "existing_dir"]),
 ]
 
--- a/src/diffpy/labpdfproc/tests/test_tools.py
+++ b/src/diffpy/labpdfproc/tests/test_tools.py
@@ -28,8 +28,7 @@ def test_set_output_directory(inputs, expected):
 
 def test_set_output_directory_bad():
     existing_file = Path().cwd().resolve() / "existing_file.py"
-    with open(existing_file, "w"):
-        pass
+    existing_file.touch()
 
     actual_args = argparse.Namespace(output_directory="existing_file.py")
     with pytest.raises(FileExistsError):
--- a/src/diffpy/labpdfproc/tests/test_tools.py
+++ b/src/diffpy/labpdfproc/tests/test_tools.py
@@ -15,5 +15,11 @@ params1 = [
 def test_set_output_directory(inputs, expected):
     expected_output_directory = expected[0]
     actual_args = argparse.Namespace(output_directory=inputs[0])
-    actual_output_directory = set_output_directory(actual_args)
-    assert actual_output_directory == expected_output_directory
+    actual_args.output_directory = set_output_directory(actual_args)
+    assert actual_args.output_directory == expected_output_directory
+    assert Path(actual_args.output_directory).exists()
+    assert Path(actual_args.output_directory).is_dir()
+
+    with pytest.raises(FileExistsError):
+        actual_args = argparse.Namespace(output_directory="test_tools.py")
+        actual_args.output_directory = set_output_directory(actual_args)
--- a/src/diffpy/labpdfproc/tests/test_tools.py
+++ b/src/diffpy/labpdfproc/tests/test_tools.py
@@ -1,0 +1,19 @@
+import argparse
+from pathlib import Path
+
+import pytest
+
+from diffpy.labpdfproc.tools import set_output_directory
+
+params1 = [
+    ([None], [Path.cwd()]),
+    (["corrected"], [Path("corrected").resolve()]),
+]
+
+
+@pytest.mark.parametrize("inputs, expected", params1)
+def test_set_output_directory(inputs, expected):
+    expected_output_directory = expected[0]
+    actual_args = argparse.Namespace(output_directory=inputs[0])
+    actual_output_directory = set_output_directory(actual_args)
+    assert actual_output_directory == expected_output_directory
--- a/src/diffpy/labpdfproc/tests/test_tools.py
+++ b/src/diffpy/labpdfproc/tests/test_tools.py
@@ -28,11 +28,11 @@ def test_set_output_directory(inputs, expected):
 
 
 def test_set_output_directory_bad():
-    existing_dir = Path().cwd().resolve() / "existing_dir.py"
-    with open(existing_dir, "w"):
+    existing_file = Path().cwd().resolve() / "existing_file.py"
+    with open(existing_file, "w"):
         pass
 
-    actual_args = argparse.Namespace(output_directory="existing_dir.py")
+    actual_args = argparse.Namespace(output_directory="existing_file.py")
     with pytest.raises(FileExistsError):
         actual_args.output_directory = set_output_directory(actual_args)
         assert Path(actual_args.output_directory).exists()
--- a/src/diffpy/labpdfproc/tests/test_tools.py
+++ b/src/diffpy/labpdfproc/tests/test_tools.py
@@ -6,13 +6,19 @@ import pytest
 from diffpy.labpdfproc.tools import set_output_directory
 
 params1 = [
-    ([None], [Path.cwd()]),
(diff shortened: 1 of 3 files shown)


sbillinge (commented):
this looks good.

Please check that the help of the arg is very clear about the behavior (i.e., read the help message and the default behavior and edit it so it is clear to the user)

On balance, I think that the default should be to write outputs in the current directory if no output dir is given.  This is probably more standard.

Also, I think it is better to work with absolute paths, so I suggest to build the full path and then load this back into the `args.output_directory` so it gets loaded correctly into metadata later.

yucongalicechen:
I updated output_dir.

For the help message, there is one comment from the description of the issue: "Files exists/force behavior updated for this eventuality." I'm not completely sure how to phrase this, or if we only need to put this for the help message for -f?

For the function, I assume that it should define and return output_dir, mkdir as needed, and maybe also load into args.output_directory for convenience? Is it helpful that I create another python file for storing this function, in addition to the functions.py we have currently for computing cve? Something like user_functions.

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -0,0 +1,19 @@
...
+    (["corrected"], [Path("corrected").resolve()]),
+]
+
+
+@pytest.mark.parametrize("inputs, expected", params1)
+def test_set_output_directory(inputs, expected):
+    expected_output_directory = expected[0]
+    actual_args = argparse.Namespace(output_directory=inputs[0])
+    actual_output_directory = set_output_directory(actual_args)
+    assert actual_output_directory == expected_output_directory
```
sbillinge, line 19: do we need to test the case where that directory alread exists and the case where it doesn't exist and make sure they both yield the right behavior?  So this function's job is to:
1) correctly determine the full path of the output directory depending on what the user provides
2) check if it exists and is a valid directory and if not create it
3) if it exists but is not a directory (a file for instance), throw an error.
4) load the full path into args.output_directory

I think the function is mostly doing this but the tests are not testing all these possibilities.  It probably means adding a few more asserts into the test function?

src/diffpy/labpdfproc/tools.py
```diff
@@ -0,0 +1,7 @@
+from pathlib import Path
+
+
+def set_output_directory(args):
```
sbillinge, line 4: If this is a public function it needs a docstring.  If a private function it needs an underscore in front. I am not sure which, but am leaning towards public as we import it from tools.

src/diffpy/labpdfproc/tests/test_tools.py
yucongalicechen: Yes I agree, I will test them.

yucongalicechen: - I think the test for cwd is a case where the directory already exists, and the other case "corrected" is where it doesn't exist. I added two assert functions for checking the actual output directory exists and is a valid directory.
- If the output directory is an existing file, mkdir will raise a FileExistsError. Do we still need to provide an error message for this case? I prefer not to because I think the current error message is clear enough. I tried to write tests for this case too, but it fails, and I would probably also need some advice on this.
- If the user provides a directory like "text.txt" but there is no existing file with this name, I believe we should still let it create the directory instead of making it an error? So that this function is easier to use as there are not too many constraints for users.
- For the function itself, I assign the output_dir computed from the function to args.output_directory, instead of writing it inside the function.

src/diffpy/labpdfproc/tools.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+    set the output directory based on the given input arguments
+
+    Parameters
+    ----------
+    args argparse.Namespace
+        input arguments provided by the user
```
sbillinge, line 11: "the arguments from the parser"

sbillinge (commented):
I think we still need more tests:
```
    ([None], [Path.cwd().resolve()]),    # tests no path provided by user
    (["."], [Path.cwd().resolve()]),    # tests no "." provided by user who wants to explicitly state cwd
    (["new_dir"], [Path().cwd().resolve() / "new_dir"]),  # tests creation of new directory specified by user
    (["existing_dir"], [Path().cwd().resolve() / "existing_dir"]),  # tests case where users specifies an existing directory (you will have to have your testing code create this dir before calling the function)
```

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -0,0 +1,25 @@
...
+@pytest.mark.parametrize("inputs, expected", params1)
+def test_set_output_directory(inputs, expected):
+    expected_output_directory = expected[0]
+    actual_args = argparse.Namespace(output_directory=inputs[0])
+    actual_args.output_directory = set_output_directory(actual_args)
+    assert actual_args.output_directory == expected_output_directory
+    assert Path(actual_args.output_directory).exists()
+    assert Path(actual_args.output_directory).is_dir()
+
+    with pytest.raises(FileExistsError):
```
sbillinge, line 23: I think the code is easier to read if you make a new function called `test_set_output_directory_bad()` and then test the cases that should fail.  Then to make it easy to read, create the file in the test and then test that it the exception is raised by the function.

I wonder though if we want to catch any case where `Path().exists()` returns true and `Path().is_dir()` returns false?  Isn't that better?

sbillinge (commented):
Please see this review but also the earlier one.

src/diffpy/labpdfproc/tests/test_tools.py
yucongalicechen: I added the tests. I think the case where Path().exists() is true and Path().is_dir() is false refers to the situation when user provides an existing file directory, so I added it in the bad case. On the other hand, if user writes the output directory as new_dir.py, I count it as one of the acceptable cases (the function creates a directory called new_dir.py). Please let me know if this is not a good idea.

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -0,0 +1,39 @@
...
+
+import pytest
+
+from diffpy.labpdfproc.tools import set_output_directory
+
+params1 = [
+    ([None], [Path.cwd().resolve()]),
+    (["."], [Path.cwd().resolve()]),
+    (["new_dir"], [Path.cwd().resolve() / "new_dir"]),
+    (["new_dir.py"], [Path.cwd().resolve() / "new_dir.py"]),
```
sbillinge, line 12: I think this example is confusing so I would just delete it.  It is also testing the same thing as the line above, unless a file exists with that name and then it is tested below.  

summary: delete

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -0,0 +1,39 @@
...
+    actual_args = argparse.Namespace(output_directory=inputs[0])
+    actual_args.output_directory = set_output_directory(actual_args)
+    assert actual_args.output_directory == expected_output_directory
+    assert Path(actual_args.output_directory).exists()
+    assert Path(actual_args.output_directory).is_dir()
+
+
+def test_set_output_directory_bad():
+    existing_file = Path().cwd().resolve() / "existing_file.py"
+    with open(existing_file, "w"):
```
sbillinge, line 32: I think `existing_file.touch()` is maybe a bit more elegant?

src/diffpy/labpdfproc/tests/test_tools.py
sbillinge: That is certainly one situation it covers, but I think anything that causes a crash is already covered.  We only need to handle the case where we are handling an Exception (not letting it crash) or some other "magic" situation where the code could behave unexpectedly, for example, the users requests an output directory but the code doesn't create it for some reason that doesn't cause a crash and then it writes in the current directory, or sthg.

So testing for "Path().exists() returns true and Path().is_dir()" and raising something like a RunTimeError with a helpful error message if this fails.  To test it we would have to  come up with a way of accessing that situation.  I can't think of any, but having that in the code makes me feel more secure because just because I can't think how to do it doesn't mean it can't be done.....

sbillinge (commented):
please see comments.  Nearly there.

yucongalicechen:
> please see comments. Nearly there.

sbillinge:
replaced by #39

PR #35, opened 2024-04-29
closes #24: check if args.input_file is specified

### LINKED ISSUE #24: check if args.input_file is specified
check if args.input_file is specified. If not, glob the inputs directory for all readable files and correct them. skip non-readable (by loadData (or DO.load when it is written). Write outputs to output directory check if files are there or not and whether force is set. Only compute cve one time and then apply it to all the files in teh glob list.

yucongalicechen:
Initial commit

Specify input directory with default as cwd, glob all files in directory and write outputs
Skip non-readable files using loadData (try except and provide an error message)

(need suggestions) still need to implement the following: compute cve only one time. We probably need to split compute_cve function to achieve this, having one function compute cves using tth we defined, and a second function to interpolat across different tths for different input files.

### DIFF
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -90,7 +90,12 @@ def main():
             )
 
         input_pattern = Diffraction_object(wavelength=wavelength)
-        xarray, yarray = loadData(filepath.name, unpack=True)
+        try:
+            xarray, yarray = loadData(filepath, unpack=True)
+        except Exception as e:
+            print(f"Non-readable data file {filepath}: {e}")
+            continue
+
         input_pattern.insert_scattering_quantity(
             xarray,
             yarray,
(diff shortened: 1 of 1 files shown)


src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -90,7 +90,12 @@ def main():
             )
 
         input_pattern = Diffraction_object(wavelength=wavelength)
-        xarray, yarray = loadData(filepath.name, unpack=True)
+        try:
+            xarray, yarray = loadData(filepath, unpack=True)
+        except Exception as e:
```
sbillinge, line 95: please specify a specific exception or exceptions, not a general one.

sbillinge (commented):
please see inline comment.  What are some read failures?

sbillinge:
This is supposed to link to the open PR that replaced it, but it is not a
big deal, don't bother updating it. For future reference, I would write
something like "closing as replaced by #50" something like that.

PR #36, opened 2024-04-29
closes #25: overload wavelength

### LINKED ISSUE #25: overload wavelength
user can specify wavelength. If it is specified have it overload the wavelength pulled from the WAVELENGTHS set

yucongalicechen:
Initial commit for overloading wavelength

Code does the following: first set wavelength to default as anode-type Mo, if args.wavelength is provided then set it to args.wavelength, otherwise if anode-type is provided set it to that.

I can wrap this up into a function if we prefer that

### DIFF
--- a/src/diffpy/labpdfproc/tests/test_tools.py
+++ b/src/diffpy/labpdfproc/tests/test_tools.py
@@ -1,7 +1,10 @@
+from unittest.mock import MagicMock
+
 import pytest
 
-from diffpy.labpdfproc.labpdfprocapp import WAVELENGTHS
-from diffpy.labpdfproc.tools import load_wavelength
+from diffpy.labpdfproc.tools import set_wavelength
+
+WAVELENGTHS = {"Mo": 0.71, "Ag": 0.59, "Cu": 1.54}
 
 params1 = [
     ([None, None], [0.71]),
@@ -12,7 +15,9 @@ params1 = [
 
 
 @pytest.mark.parametrize("inputs, expected", params1)
-def test_load_wavelengths(inputs, expected):
+def test_set_wavelength(inputs, expected):
     expected_wavelength = expected[0]
-    actual_wavelength = load_wavelength(inputs[0], inputs[1], WAVELENGTHS)
+    actual_args = MagicMock()
+    actual_args.wavelength, actual_args.anode_type = inputs[0], inputs[1]
+    actual_wavelength = set_wavelength(actual_args)
     assert actual_wavelength == expected_wavelength
--- a/src/diffpy/labpdfproc/tests/test_tools.py
+++ b/src/diffpy/labpdfproc/tests/test_tools.py
@@ -1,0 +1,18 @@
+import pytest
+
+from diffpy.labpdfproc.labpdfprocapp import WAVELENGTHS
+from diffpy.labpdfproc.tools import load_wavelength
+
+params1 = [
+    ([None, None], [0.71]),
+    ([None, "Ag"], [0.59]),
+    ([0.25, "Ag"], [0.25]),
+    ([0.25, None], [0.25]),
+]
+
+
+@pytest.mark.parametrize("inputs, expected", params1)
+def test_load_wavelengths(inputs, expected):
+    expected_wavelength = expected[0]
+    actual_wavelength = load_wavelength(inputs[0], inputs[1], WAVELENGTHS)
+    assert actual_wavelength == expected_wavelength
--- a/src/diffpy/labpdfproc/tests/test_tools.py
+++ b/src/diffpy/labpdfproc/tests/test_tools.py
@@ -20,10 +20,15 @@ def test_set_wavelength(inputs, expected):
     assert actual_wavelength == expected_wavelength
 
 
-def test_set_wavelength_bad():
+params3 = [
+    ([None, "invalid"]),
+    ([0, None]),
+    ([-1, "Mo"]),
+]
+
+
+@pytest.mark.parametrize("inputs", params3)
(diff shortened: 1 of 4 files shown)


sbillinge (commented):
Yes, please wrap in a function with tests.

sbillinge (commented):
please see inline comment.

Maybe call the module `tools.py`

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+
+from diffpy.labpdfproc.labpdfprocapp import WAVELENGTHS
+from diffpy.labpdfproc.tools import load_wavelength
+
+params1 = [
+    ([None, None], [0.71]),
+    ([None, "Ag"], [0.59]),
```
sbillinge, line 8: maybe also test an unknown string?

src/diffpy/labpdfproc/tools.py
```diff
@@ -0,0 +1,7 @@
+def load_wavelength(args_wavelength, args_anode_type, WAVELENGTHS):
```
sbillinge, line 1: I suggest to not pass in `WAVELENGTHS`.  define this as a global in tools.

Is there a reason that you don't just pass in `args` and then return `args`?  I think this would be more elegant.

sbillinge (commented):
please see comments

src/diffpy/labpdfproc/tools.py
yucongalicechen: I think it's easier to write tests if we pass them separately? since we just need 2 parameters.

src/diffpy/labpdfproc/tests/test_tools.py
yucongalicechen: For an unknown string for anode_type, the function should still return default value, right?

src/diffpy/labpdfproc/tools.py
yucongalicechen: Per your comment on another pull, "we need to load the correct wavelength into the args.wavelength variable even if this was not specified on the command line, and that i swhat the set_wavelength function will do."

We currently have the following codes related to wavelength:
wavelength = set_wavelength(...)
input_pattern = Diffraction_object(wavelength=wavelength)

Would it be better if I just replace wavelength with args.wavelength?

src/diffpy/labpdfproc/tests/test_tools.py
sbillinge: this is your decision.  What do you think the user would want?  Write the test for that.  Then implement it.

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -0,0 +1,23 @@
...
+    ([0.25, "Ag"], [0.25]),
+    ([0.25, None], [0.25]),
+]
+
+
+@pytest.mark.parametrize("inputs, expected", params1)
+def test_set_wavelength(inputs, expected):
+    expected_wavelength = expected[0]
+    actual_args = MagicMock()
```
sbillinge, line 20: This works, but it may be better to create an argparse namespace and populate it rather than doing a Mock.  This would be closer to the actual behavior.

sbillinge (commented):
getting there.  Please see comments.

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -1,23 +1,22 @@
-from unittest.mock import MagicMock
+import argparse
 
 import pytest
 
 from diffpy.labpdfproc.tools import set_wavelength
 
 WAVELENGTHS = {"Mo": 0.71, "Ag": 0.59, "Cu": 1.54}
```
sbillinge, line 7: delete from here.

src/diffpy/labpdfproc/tests/test_tools.py
sbillinge: I see code below but not the tests for these things....please make the tests first, then you can write the code.

src/diffpy/labpdfproc/tools.py
```diff
@@ -1,7 +1,17 @@
...
 WAVELENGTHS = {"Mo": 0.71, "Ag": 0.59, "Cu": 1.54}
 
 
 def set_wavelength(args):
+    if args.wavelength and not isinstance(args.wavelength, float):
+        sys.exit("Please enter the wavelength as a float value.")
+    if not args.wavelength and args.anode_type and not (args.anode_type in WAVELENGTHS):
+        sys.exit(
+            "Please either specify a wavelength as a float value "
+            "or specify anode_type as one of 'Mo', 'Ag', or 'Cu'."
```
sbillinge, line 12: use an f-string and unpack WAVELENGTHS so when we update wavelengths the warning message updates automatically.

Also specify it as an error and tell teh user what to do:  "please rerun the program specifying...."  The best may be to throw a `ValueError` Exception rather than sys.exit which is not good practice.

sbillinge (commented):
please see inline comments.

src/diffpy/labpdfproc/tools.py
yucongalicechen: Edited. I also added tests for invalid anode types and non-positive wavelengths. I find these two lines

WAVELENGTHS = {"Mo": 0.71, "Ag": 0.59, "Cu": 1.54}
known_sources = [key for key in WAVELENGTHS.keys()]

appear in both labpdfprocapp.py and tools.py, but we need known_sources in the help message of anode type. Maybe we should just move them to tools.py?

PR #37, opened 2024-04-30
closes #26: initial commit on loading user-specified pairs into args

### LINKED ISSUE #26: insert user-specified key-value pairs into DO.metadata
allow user to specify key-value pairs. Insert these into DO.metadata. do nothing with them, but they will be written into the header of the output chi files.

yucongalicechen:
Initial commit on loading extra pairs into args, tests not written yet.

The pairs are not written into the header of output files for now, as required in the issue description, but this functionality should be addressed from https://github.com/diffpy/diffpy.labpdfproc/issues/27?

### DIFF
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -3,7 +3,7 @@ from argparse import ArgumentParser
 from pathlib import Path
 
 from diffpy.labpdfproc.functions import apply_corr, compute_cve
-from diffpy.labpdfproc.tools import load_additional_info
+from diffpy.labpdfproc.tools import load_user_metadata
 from diffpy.utils.parsers.loaddata import loadData
 from diffpy.utils.scattering_objects.diffraction_objects import XQUANTITIES, Diffraction_object
 
@@ -37,7 +37,7 @@ def get_args():
     p.add_argument(
         "-x",
         "--xtype",
-        help=f"the quantity on the independnt variable axis. allowed "
+        help=f"the quantity on the independent variable axis. allowed "
         f"values: {*XQUANTITIES, }. If not specified then two-theta "
         f"is assumed for the independent variable. Only implemented for "
         f"tth currently",
@@ -55,15 +55,15 @@ def get_args():
         "-f",
         "--force-overwrite",
         action="store_true",
-        help="outputs will not overwrite existing file unless --force is spacified",
+        help="outputs will not overwrite existing file unless --force is specified",
     )
     p.add_argument(
-        "-add",
-        "--additional-info",
+        "-u",
+        "--user-metadata",
         metavar=("KEY=VALUE"),
         action="append",
         help="specify key-value pairs to be loaded into metadata by using key=value. "
-        "You can specify multiple paris by calling -add multiple times.",
+        "You can specify multiple paris by calling -u multiple times.",
     )
     args = p.parse_args()
     return args
@@ -71,7 +71,7 @@ def get_args():
 
 def main():
     args = get_args()
-    args = load_additional_info(args)
+    args = load_user_metadata(args)
     wavelength = WAVELENGTHS[args.anode_type]
     filepath = Path(args.input_file)
     outfilestem = filepath.stem + "_corrected"
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -3,6 +3,7 @@ from argparse import ArgumentParser
 from pathlib import Path
 
 from diffpy.labpdfproc.functions import apply_corr, compute_cve
+from diffpy.labpdfproc.tools import load_additional_info
 from diffpy.utils.parsers.loaddata import loadData
 from diffpy.utils.scattering_objects.diffraction_objects import XQUANTITIES, Diffraction_object
 
@@ -56,12 +57,21 @@ def get_args():
         action="store_true",
         help="outputs will not overwrite existing file unless --force is spacified",
     )
+    p.add_argument(
+        "-add",
+        "--additional-info",
+        metavar=("KEY=VALUE"),
+        action="append",
+        help="specify key-value pairs to be loaded into metadata. You can specify multiple "
+        "paris by calling -add multiple times",
+    )
     args = p.parse_args()
     return args
 
 
 def main():
     args = get_args()
+    args = load_additional_info(args)
     wavelength = WAVELENGTHS[args.anode_type]
     filepath = Path(args.input_file)
     outfilestem = filepath.stem + "_corrected"
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -62,8 +62,8 @@ def get_args():
         "--additional-info",
         metavar=("KEY=VALUE"),
         action="append",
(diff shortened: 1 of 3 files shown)


src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -56,12 +57,21 @@ def get_args():
         action="store_true",
         help="outputs will not overwrite existing file unless --force is spacified",
     )
+    p.add_argument(
+        "-add",
```
sbillinge, line 61: I would call this `--user-metadata` and `-u` if that is available.

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -56,12 +57,21 @@ def get_args():
         action="store_true",
         help="outputs will not overwrite existing file unless --force is spacified",
```
sbillinge, line 58: please correct typo in this line, I just saw it.

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -0,0 +1,30 @@
...
+
+@pytest.mark.parametrize("inputs, expected", params5)
+def test_load_additional_info(inputs, expected):
+    actual_parser = ArgumentParser()
+    actual_parser.add_argument("-add", "--additional-info", action="append", metavar=("KEY=VALUE"))
+    actual_args = actual_parser.parse_args([])
+    expected_parser = ArgumentParser()
+    expected_args = expected_parser.parse_args([])
+
+    if not inputs:
```
sbillinge, line 22: I am not sure why you have conditionals in the test.  Isn't the test that you pass whatever the inputs are to the function and then the function should handle things as you want?

sbillinge (commented):
I am not sure if this will work as intended.  How do I enter something that would result in
{"toast": "for breakfast"} 
or
{"mylist": [1,2,3.0]}
and so on.

src/diffpy/labpdfproc/tests/test_tools.py
yucongalicechen: Yes thanks for catching that. I changed the codes.

yucongalicechen:
> I am not sure if this will work as intended. How do I enter something that would result in {"toast": "for breakfast"} or {"mylist": [1,2,3.0]} and so on.

I added the tests for these two cases. I think this works by using "=". But if user needs to have more than one equals sign, this might be a problem. Like {"data for n=1": [1,2,3]}. Would we need to consider these cases?
I also saw your other comments on output_directory. I assume we prefer throwing an error here instead of using sys.exit, when encountering an invalid key-value pair?

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -0,0 +1,27 @@
...
+    ([[]], []),
+    ([["toast=for breakfast"]], [["toast", "for breakfast"]]),
+    ([["mylist=[1,2,3.0]"]], [["mylist", "[1,2,3.0]"]]),
+    ([["weather=rainy", "day=tuesday"]], [["weather", "rainy"], ["day", "tuesday"]]),
+]
+
+
+@pytest.mark.parametrize("inputs, expected", params5)
+def test_load_additional_info(inputs, expected):
+    actual_parser = ArgumentParser()
```
sbillinge, line 17: Since this one is a bit more tricky, I htink it is better if we do actually test the parser in this case, so actually run the parser and don't just create a new namespace.

For example, i know htat your "toast", "for breakfast" passes the tests, but what do I actually type on the command line to make that work?  Do I type `labpdfproc --u toast=for breakfast` or do I have to type `labpdfproc --u toast="for breakfast"`

On the whole, if we are going to go with the approach of parsing strings and looking for "=" it would be better if we used `nargs='+'` and allowed multiple key-value pairs to be passed at the same time (this feels more intuitive to me as a user), then allow the user to escape the = with a slash, for example.  We then need to give very good instructions in the help, e.g., see here https://stackoverflow.com/questions/27146262/create-variable-key-value-pairs-with-argparse-python

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -0,0 +1,27 @@
+from argparse import ArgumentParser
+
+import pytest
+
+from diffpy.labpdfproc.tools import load_user_metadata
+
+params5 = [
+    ([[]], []),
+    ([["toast=for breakfast"]], [["toast", "for breakfast"]]),
+    ([["mylist=[1,2,3.0]"]], [["mylist", "[1,2,3.0]"]]),
```
sbillinge, line 10: I think this is ok since we are just writing the metadata to a file, but actually what i had in mind for the output was `["mylist", [1,2,3.0]]` not `["mylist", "[1,2,3.0]"]`.

sbillinge (commented):
This is a tricky one but we are making progress.  Please see the inline comments.  Maybe start with the lowest one and work back....

sbillinge:
> Initial commit on loading extra pairs into args, tests not written yet.
> 
> The pairs are not written into the header of output files for now, as required in the issue description, but this functionality should be addressed from #27?

They will be written to the file by the `Diffraction_object.dump()`.  It would be good to check that this is working at that time.

src/diffpy/labpdfproc/tools.py
```diff
@@ -34,3 +34,36 @@ def set_wavelength(args):
         return WAVELENGTHS[args.anode_type]
     else:
         return WAVELENGTHS["Mo"]
+
+
+def _load_key_value_pair(s):
+    items = s.split("=")
+    key = items[0].strip()
+    if len(items) > 1:
+        value = "=".join(items[1:])
```
sbillinge, line 43: this is making some assumptions about which is the "splitting" = sign.  This is ok, but let's let the people know what hte rule is that we are applying (equals sign not allowed in key name?) in the help message to the argparse.  I htink adding a few examples to the help might be good too, but don't do toast for breakfast but something topical, like `-u "facility=NSLS II" beamline=28ID-2`

src/diffpy/labpdfproc/tools.py
```diff
@@ -34,3 +34,36 @@ def set_wavelength(args):
...
+        for item in args.user_metadata:
+            if "=" not in item:
+                raise ValueError("please provide key-value pairs in the format key=value.")
+            key, value = _load_key_value_pair(item)
+            setattr(args, key, value)
+    delattr(args, "user_metadata")
+    return args
+
+
+# Notes, can ignore
```
sbillinge, line 58: this alll needs to be in a docstring or in the help message (but as I said above, better to give examples), and represented in the tests.  Then delete these comments

sbillinge (commented):
please see inline comments

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -55,15 +55,26 @@ def get_args():
         "-f",
         "--force-overwrite",
         action="store_true",
-        help="outputs will not overwrite existing files unless --force is spacified",
+        help="outputs will not overwrite existing file unless --force is specified",
```
sbillinge, line 58: I think I have seen this fix on multiple branches (not sure) but only fix things in one branch or your generate conflicts (as we have been finding)

sbillinge (commented):
please see inline comments

sbillinge (commented):
please see inline comments.  I think something has gone wrong with a merge.  It is probably better to close this PR and blow away the branch and do it over on a clean branch built from a fully synced main, hten reapply the edits

PR #40, opened 2024-05-02
closes #25: update wavelength function

### LINKED ISSUE #25: overload wavelength

yucongalicechen:
We refer {*known_sources, } which gives the output ('Mo', ...), do we prefer ['Mo', ...] or do we not have a preference about this?

### DIFF
--- a/src/diffpy/labpdfproc/tests/test_tools.py
+++ b/src/diffpy/labpdfproc/tests/test_tools.py
@@ -1,4 +1,5 @@
 import argparse
+import re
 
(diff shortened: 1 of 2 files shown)


sbillinge:
Thanks so much @yucongalicechen .   regarding the form of the error message,  I think this doesn't matter too much.  Let's see if the current message causes confusion or not when it is used.

PR #41, opened 2024-05-02
closes #22: set tmppath for output dir

### LINKED ISSUE #22: specify an output directory name

yucongalicechen:


src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -1,9 +1,49 @@
...
+    assert actual_args.output_directory == expected_output_directory
+    assert Path(actual_args.output_directory).exists()
+    assert Path(actual_args.output_directory).is_dir()
+
+
+def test_set_output_directory_bad(tmp_path):
+    directory = Path(tmp_path)
+    os.chdir(directory)
+
+    existing_file = Path().cwd().resolve() / "existing_file.py"
```
sbillinge, line 38: To make it less confusing let's just set this to `Path(tmp_path).resolve()` and not `Path().cwd().resolve()`

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -30,14 +30,15 @@ def get_args():
...
+        help="the name of the output directory. If not specified "
+        "then corrected files will be written to the current directory."
+        "If the specified directory doesn't exist it will be created.",
         default=None,
     )
     p.add_argument(
         "-x",
         "--xtype",
-        help=f"the quantity on the independnt variable axis. allowed "
+        help=f"the quantity on the independent variable axis. allowed "
```
sbillinge, line 41: Let's capitalize all the first letters of all the help statements so it is better grammar.

sbillinge (commented):
please see inline comments.

sbillinge:
Yay @yucongalicechen

PR #42, opened 2024-05-03
update fixture

### LINKED ISSUE #23: specify an inputs directory name
specify an inputs directory name. Default is cwd.

yucongalicechen:
Requirements:
- specify an inputs directory name. Default is cwd.
- check if args.input_file is specified. If not, glob the inputs directory for all readable files and correct them. skip non-readable (by loadData (or DO.load when it is written). Write outputs to output directory check if files are there or not and whether force is set. Only compute cve one time and then apply it to all the files in the glob list.

Current implements:
- set_input_directory function based on input file argument
- glob all files in input directory. for every input file, use try except, raise value error with error message
- implemented user_filesystem in conftest to build tmp_dir for tests (updated tests accordingly)
- tests for input_directory for no user input and user input valid files. put an absolute path case and commented it.

Other things to implement based on discussion (maybe open issues?):
- tests for valid/invalid input files: would need some advice on this, since current code decides valid/invalid input files based on loadData function. The current input_directory can only test if input file exists or not.
- seperate compute_cve function on another branch
- functionality for user to input a list of files / directories

### DIFF
--- a/src/diffpy/labpdfproc/tests/conftest.py
+++ b/src/diffpy/labpdfproc/tests/conftest.py
@@ -9,22 +9,35 @@ def user_filesystem(tmp_path):
     directory = Path(tmp_path)
     os.chdir(directory)
 
-    existing_dir = Path(tmp_path).resolve() / "existing_dir"
-    existing_dir.mkdir(parents=True, exist_ok=True)
+    input_dir = Path(tmp_path).resolve() / "input_dir"
+    input_dir.mkdir(parents=True, exist_ok=True)
 
-    existing_file = Path(tmp_path).resolve() / "existing_file.py"
-    existing_file.touch()
+    chi_data = "dataformat = twotheta\n mode = xray\n # chi_Q chi_I\n 1 2\n 3 4\n 5 6\n 7 8\n"
+    xy_data = "1 2\n 3 4\n 5 6\n 7 8"
+    txt_data = "dataformat = twotheta\n mode = xray\n # chi_Q chi_I\n 1 2\n 3 4\n 5 6\n 7 8\n"
+    unreadable_data = "This is an unreadable file."
+    binary_data = b"\x00\x01\x02\x03\x04"
 
-    input_directory = Path(tmp_path).resolve() / "input_directory"
-    input_directory.mkdir(parents=True, exist_ok=True)
+    with open("good_data.chi", "w") as f:
+        f.write(chi_data)
+    with open("good_data.xy", "w") as f:
+        f.write(xy_data)
+    with open("good_data.txt", "w") as f:
+        f.write(txt_data)
+    with open("unreadable_file.txt", "w") as f:
+        f.write(unreadable_data)
+    with open("binary.pkl", "wb") as f:
+        f.write(binary_data)
 
-    with open(os.path.join(input_directory, "good_data.chi"), "w") as f:
-        f.write("1 2 \n 3 4 \n 5 6 \n 7 8")
-    with open(os.path.join(input_directory, "good_data.xy"), "w") as f:
-        f.write("1 2 \n 3 4 \n 5 6 \n 7 8")
-    with open(os.path.join(input_directory, "unreadable_file.txt"), "w") as f:
-        f.write("This is an unreadable file.")
-    with open(os.path.join(input_directory, "binary.pkl"), "wb") as f:
-        f.write(b"\x00\x01\x02\x03\x04")
+    with open(os.path.join(input_dir, "good_data.chi"), "w") as f:
+        f.write(chi_data)
+    with open(os.path.join(input_dir, "good_data.xy"), "w") as f:
+        f.write(xy_data)
+    with open(os.path.join(input_dir, "good_data.txt"), "w") as f:
+        f.write(txt_data)
+    with open(os.path.join(input_dir, "unreadable_file.txt"), "w") as f:
+        f.write(unreadable_data)
+    with open(os.path.join(input_dir, "binary.pkl"), "wb") as f:
+        f.write(binary_data)
 
     yield tmp_path
--- a/src/diffpy/labpdfproc/tests/conftest.py
+++ b/src/diffpy/labpdfproc/tests/conftest.py
@@ -1,0 +1,18 @@
+import os
+from pathlib import Path
+
+import pytest
+
+
+@pytest.fixture
+def user_filesystem(tmp_path):
+    directory = Path(tmp_path)
+    os.chdir(directory)
+
+    existing_dir = Path(tmp_path).resolve() / "existing_dir"
+    existing_dir.mkdir(parents=True, exist_ok=True)
+
+    existing_file = Path(tmp_path).resolve() / "existing_file.py"
+    existing_file.touch()
+
+    yield tmp_path
--- a/src/diffpy/labpdfproc/tests/conftest.py
+++ b/src/diffpy/labpdfproc/tests/conftest.py
@@ -14,7 +14,6 @@ def user_filesystem(tmp_path):
 
     chi_data = "dataformat = twotheta\n mode = xray\n # chi_Q chi_I\n 1 2\n 3 4\n 5 6\n 7 8\n"
     xy_data = "1 2\n 3 4\n 5 6\n 7 8"
-    txt_data = "dataformat = twotheta\n mode = xray\n # chi_Q chi_I\n 1 2\n 3 4\n 5 6\n 7 8\n"
     unreadable_data = "This is an unreadable file."
     binary_data = b"\x00\x01\x02\x03\x04"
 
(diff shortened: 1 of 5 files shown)


sbillinge (commented):
please see inline comments.  I don't think we need special handling of badly formulated files so we probably don't need any `_bad` tests.

Also, I think we don't need the re.escape() but let me know if there is a reason you need it.

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -57,12 +57,37 @@ def test_set_input_directory(inputs, expected, tmp_path):
...
+
     expected_input_directory = Path(tmp_path).resolve() / expected[0]
     actual_args = argparse.Namespace(input_file=inputs[0])
     actual_args = set_input_directory(actual_args)
     assert actual_args.input_directory == expected_input_directory
 
 
+params3 = [
+    (
+        ["new_dir/data.xy"],
```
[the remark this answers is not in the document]
yucongalicechen: Okay this makes sense, I'll delete this

sbillinge (commented):
I am thinking about this issue and I think the idea is that the user specifies an  input and we handle it.  The possibles I can think of are: User specifies a

1.  file.  We read the file
2. directory.  We glob the directory and read every file we can
3. a wild-card like path/to/input_dir/*.chi`.  We load all the files that satisfy that pattern

For now, if we just implement 1 and 2 and leave 3 as an issue, I think this will be fine.

yucongalicechen:
I added the issue. Changed tests from Namespace to parser for output_directory and input_directory tests.
I didn't manage to change for wavelength because we have inputs None or a float. When I pass None it's causing a typeerror, and I haven't found any way to fix this yet..

sbillinge:
👍

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -38,13 +41,40 @@ def test_set_output_directory_bad(tmp_path):
...
         actual_args.output_directory = set_output_directory(actual_args)
         assert Path(actual_args.output_directory).exists()
         assert not Path(actual_args.output_directory).is_dir()
 
 
+params2 = [
+    ([None], ["."]),
+    (["data.xy"], ["."]),
+    (["existing_dir/data.xy"], ["existing_dir"]),
+]
```
sbillinge, line 58: lets also test the user putting in an absolute file-path.  I am not sure how to test that unless there is a file path that we know will  exist on the testing computers.  If we can't figure it out,  maybe put it in but comment it out, so we know the "intent", but make sure that the code would pass if this were passed.

Also, this doesn't seem to be testing the logic that the user specifies an input file, only an input directory.  We want to support both.

sbillinge (commented):
please see inline comments

src/diffpy/labpdfproc/tests/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+
+@pytest.fixture
+def user_filesystem(tmp_path):
+    directory = Path(tmp_path)
+    os.chdir(directory)
+
+    existing_dir = Path(tmp_path).resolve() / "existing_dir"
+    existing_dir.mkdir(parents=True, exist_ok=True)
+
+    existing_file = Path(tmp_path).resolve() / "existing_file.py"
```
sbillinge, line 15: We need the following files:
```
directory/good_data.chi   # a valid chi file
directory/good_data.xy   # valid data in xy format
directory/unreadable_file.txt  # a text file that will raise an error
directory/binary.pckl             # some kind of unreadable binary file, maybe some kind of pickle file or sthg
```
then all these same files written in another directory called `input_directory`

To build this we could have the files saved in some directory in tests and then copy them to `tmp_dir` as part of the fixture, or we could write the by saving the contents in a py file somewhere and doing writes.  The data only needs to be four lines long or so, but valid format so that loadData will read it correctly.

sbillinge (commented):
please see inline comment

src/diffpy/labpdfproc/tests/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+    existing_file.touch()
+
+    input_directory = Path(tmp_path).resolve() / "input_directory"
+    input_directory.mkdir(parents=True, exist_ok=True)
+
+    with open(os.path.join(input_directory, "good_data.chi"), "w") as f:
+        f.write("1 2 \n 3 4 \n 5 6 \n 7 8")
+    with open(os.path.join(input_directory, "good_data.xy"), "w") as f:
```
sbillinge, line 23: this seems to be the same file contents as `good_data.chi`.  Actually we probably don't need this for testing, but it may be helpful for documentation.   However, the distinction in general is that chi files can have a header, then a data-block, so to make it clearer to people the intention of a test, leave this xy file as is but make the chi file look more like a real chi file?

src/diffpy/labpdfproc/tests/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+        f.write("1 2 \n 3 4 \n 5 6 \n 7 8")
+    with open(os.path.join(input_directory, "unreadable_file.txt"), "w") as f:
+        f.write("This is an unreadable file.")
```
sbillinge, line 26: `loadData` can read text headers so it may be able to read this.  Did you check that it raises an error if there is no numeric block (I actually don't know the answer to that...just make sure).

src/diffpy/labpdfproc/tests/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+
+import pytest
+
+
+@pytest.fixture
+def user_filesystem(tmp_path):
+    directory = Path(tmp_path)
+    os.chdir(directory)
+
+    existing_dir = Path(tmp_path).resolve() / "existing_dir"
```
sbillinge, line 12: since you are making the full stack of files on an `input_directory` we don't need the `existing_dir` and `existing_file` any more.

sbillinge (commented):
I am only reviewing the fixture here, so am concentrating on that till it is finished, then I can switch to re-looking at the code.  But this is looking great.  There are a few small comments, please see inline.  Also, we need those four files also written directly on `tmp_path`, so the same files are in two places, in `tmp_path` and in `tmp_path/input_dir`.  This allows us to test two different user behaviors.  The first behavior is where the user wants to just always `cd` to the directory where the files are and run the program where everything is in `cwd`.  The second behavior that we want to support is that the user is very organized and wants to run the program from somewhere but carefully specify at runtime for the program to run in some different directory by specifying it.  I am not sure if that makes sense, but that is what we are after.

yucongalicechen:
I changed set_input_dir to set_input_files because I think it better addresses the comments above (so maybe some more comments are needed..). I list everything I changed here for convenience:
- tools: For set_input_files, I added glob into this function and put all file names in args.input_file, so that in our main program, we only need to loop args.input_file.
- conftest: Deleted existing_dir and existing_file.py as requested. Added all data files in both cwd and input_dir. Added header in chi and txt files.
- test_tools: For test_output_directory I changed the existing directories and files. For test_set_input_files, there are three cases (no inputs, input file in cwd, input file in another directory). test_loadData_with_input_files edited.

src/diffpy/labpdfproc/tests/conftest.py
```diff
@@ -0,0 +1,43 @@
...
+def user_filesystem(tmp_path):
+    directory = Path(tmp_path)
+    os.chdir(directory)
+
+    input_dir = Path(tmp_path).resolve() / "input_dir"
+    input_dir.mkdir(parents=True, exist_ok=True)
+
+    chi_data = "dataformat = twotheta\n mode = xray\n # chi_Q chi_I\n 1 2\n 3 4\n 5 6\n 7 8\n"
+    xy_data = "1 2\n 3 4\n 5 6\n 7 8"
+    txt_data = "dataformat = twotheta\n mode = xray\n # chi_Q chi_I\n 1 2\n 3 4\n 5 6\n 7 8\n"
```
sbillinge, line 17: what is your intent with this txt_data?  It seems to be t he same as a chi file but with a different name, so what do you want to test with this?  I can think of some things but I was wondering your intent.

sbillinge (commented):
I am still just reviewing the conftest.py for now.  I think I would like to merge just that when it is finished, so we could comment out or remove the other changes to another branch (I am ok with either way).  This way we will hvae this nice universal testing fixture for testing across multiple things if we need it on other branches.

tl;dr, pls could you fix the last comment inline and comment/remove the rest and I will merge it (also delete the `closes #` comment in the PR as it now won't close anything.  Then let's have a quick convo, maybe on slack, about what UCs we want to implement and some code design to do it, then you can go back an reimplement those things.  Does it sound good?

sbillinge:
Awesome!

PR #44, opened 2024-05-04
initial commit on separating cve functions

yucongalicechen:


src/diffpy/labpdfproc/functions.py
```diff
@@ -172,28 +172,25 @@ def get_path_length(self, grid_point, angle):
...
     -------
-    the diffraction object with cve curves
+    the diffraction object with cve curves on the resampled grid
 
     it is computed as follows:
-    We first resample data and absorption correction to a more reasonable grid,
-    then calculate corresponding cve for the given mud in the resample grid
-    (since the same mu*D yields the same cve, we can assume that D/2=1, so mu=mud/2),
-    and finally interpolate cve to the original grid in diffraction_data.
+    We resample data and absorption correction to a more reasonable grid,
```
sbillinge, line 191: This is a very excellent explanation, but normally something like this that expands on what the function does, I think it is place below the first line of the docstring after a blank line, not down here at the bottom.

src/diffpy/labpdfproc/functions.py
```diff
@@ -208,7 +205,40 @@ def compute_cve(diffraction_data, mud, wavelength):
...
+
+    Returns
+    -------
+    the diffraction object with cve curves
+
+    """
+
     orig_grid = diffraction_data.on_tth[0]
+    cve = abdo_resampled.on_tth[1]
     newcve = np.interp(orig_grid, TTH_GRID, cve)
```
sbillinge, line 242: this would be easier to read if, instead of `TTH_GRID` you used `abdo_resampled.on_tth[0]`.  Actually, all in all, I think it would be easier to read as:
```
cve_on_data_grid = np.interp(diffraction_data.on_tth[0], cve.on_tth[0], cve.on_tth[1])
```

sbillinge (commented):
Before we do this we probably want the tests that implement the desired behavior.   

The first question to answer is: 
What is the desired behavior?

Then:
What is the best way to do that?

I think the basic issue is that if we glob a directory we can find all kinds of things.  If we take the route you have been taking the program will read everything it can read (and not everything else), then it will compute a cve and apply it to everything that loadData successfully loaded, regardless of what was in there.    

This is perfectly acceptable behavior, but it may be better for the program to be more constrained. Some scenarios could be that I have a bunch of files in the directory that I measured at different times on different grids from different samples and I forget that.  In that case, it might be better for the program to say "hey, Simon, I find files with different compositions and grids in here, what do you actually want me to do?  I may need the help of a human to resolve this actually", kind of like git does when it finds a conflict.   If we implemented the code as we are here, it would be kind of like git making its own best guess effort at thinking what hte user had in mind when resolving conflicts.

On the other hand, maybe this would be really frustrating for users because they measured hundreds of files from the same sample and because of some weird reason like rounding errors,  each was on a slightly different grid.  If we had the picky code, it would be very inconvenient to use.

In agile programming, we usually choose to implement the "minimum viable" solution that satisfies the UC.  We add more functionality when our early adopter users request that. In this case, the mvp would probably be to glob the directory, read everything we can read, apply the correction to everything if it is on the same grid and raise an exception if this is not the case.  The exception would be most helpful if it told the user which files were on different grids somehow and what to do to fix the situation (rerun specifying files separately, specifying a list of files [not implemented yet] or putting files into separate directories if they are on different grids.

It's not that we can't interpolate everything onto a common grid (we can) but more whether we actually want to do that.  Will it result in better science or not?

One final comment.  This was functionality that I wanted in the Diffraction_object itself, so that I could, for example, plot on top of each other two diffraction objects that were on different grids to compare them (we often run into this) but it never quite got implemented.  It would be much better for me if we implemented this behavior in Diffraction_object (so put a PR into there) with this behavior.....though I am still not convinced that we want to expose that behavior here.....

PR #47, opened 2024-05-04
closes #26: loading user-specified key-value pairs

### LINKED ISSUE #26: insert user-specified key-value pairs into DO.metadata

yucongalicechen:
- Moved everything from previous branch, including adding -u argument, function for load_user_metadata, and basic tests.
- Edited help messgae, included docstring and more tests. Tested good cases (no inputs, put quotes around whitespace, no whitespaces before and after =, specify repeated keys, pairs with more than one = sign) and bad cases (do not put quotes around whitespace). I'm not sure if the inputs for test cases are intuitive enough. For example, if I specify `-u weather=rainy day=tuesday` or `-u "weather=rainy" "day=tuesday"`, then they are both interpreted as `["weather=rainy", "day=tuesday"]` when I print args.user_metadata. So this is what I put for inputs.

### DIFF
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -3,7 +3,7 @@ from argparse import ArgumentParser
 from pathlib import Path
 
 from diffpy.labpdfproc.functions import apply_corr, compute_cve
-from diffpy.labpdfproc.tools import load_additional_info
+from diffpy.labpdfproc.tools import load_user_metadata
 from diffpy.utils.parsers.loaddata import loadData
 from diffpy.utils.scattering_objects.diffraction_objects import XQUANTITIES, Diffraction_object
 
@@ -37,7 +37,7 @@ def get_args():
     p.add_argument(
         "-x",
         "--xtype",
-        help=f"the quantity on the independnt variable axis. allowed "
+        help=f"the quantity on the independent variable axis. allowed "
         f"values: {*XQUANTITIES, }. If not specified then two-theta "
         f"is assumed for the independent variable. Only implemented for "
         f"tth currently",
@@ -55,15 +55,15 @@ def get_args():
         "-f",
         "--force-overwrite",
         action="store_true",
-        help="outputs will not overwrite existing file unless --force is spacified",
+        help="outputs will not overwrite existing file unless --force is specified",
     )
     p.add_argument(
-        "-add",
-        "--additional-info",
+        "-u",
+        "--user-metadata",
         metavar=("KEY=VALUE"),
         action="append",
         help="specify key-value pairs to be loaded into metadata by using key=value. "
-        "You can specify multiple paris by calling -add multiple times.",
+        "You can specify multiple paris by calling -u multiple times.",
     )
     args = p.parse_args()
     return args
@@ -71,7 +71,7 @@ def get_args():
 
 def main():
     args = get_args()
-    args = load_additional_info(args)
+    args = load_user_metadata(args)
     wavelength = WAVELENGTHS[args.anode_type]
     filepath = Path(args.input_file)
     outfilestem = filepath.stem + "_corrected"
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -3,6 +3,7 @@ from argparse import ArgumentParser
 from pathlib import Path
 
 from diffpy.labpdfproc.functions import apply_corr, compute_cve
+from diffpy.labpdfproc.tools import load_additional_info
 from diffpy.utils.parsers.loaddata import loadData
 from diffpy.utils.scattering_objects.diffraction_objects import XQUANTITIES, Diffraction_object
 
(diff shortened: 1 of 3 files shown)


src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -76,3 +77,44 @@ def test_set_wavelength_bad(inputs, msg):
...
         actual_args.wavelength = set_wavelength(actual_args)
+
+
+params5 = [
+    ([None], []),
+    (
+        [["facility=NSLS II", "beamline=28ID-2", "favorite color=blue"]],
+        [["facility", "NSLS II"], ["beamline", "28ID-2"], ["favorite color", "blue"]],
+    ),
+    ([["facility=NSLS II", "facility=NSLS III"]], [["facility", "NSLS III"]]),
```
sbillinge, line 88: This is a great test (I like how you are thinking here), but might it not be better to raise an error in this case?  Otherwise, like the git example we discussed before, you are having the code make decisions for the human and that breaks our magic rule.  Better to let the person know there is a boo boo and ask them to rerun fixing the boo boo (don't use the words "boo boo"!)

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -76,3 +77,44 @@ def test_set_wavelength_bad(inputs, msg):
...
+
+
+@pytest.mark.parametrize("inputs, expected", params5)
+def test_load_user_metadata(inputs, expected):
+    expected_parser = ArgumentParser()
+    expected_args = expected_parser.parse_args([])
+    for expected_pair in expected:
+        setattr(expected_args, expected_pair[0], expected_pair[1])
+
+    actual_parser = ArgumentParser()
```
sbillinge, line 100: I still think we should use the actual parser, not create a parser called `actual_parser` (sorry if I wasn't clear).   It will start with something at the top of the file that looks like
```
from diffpy.labpdfproc.labpdfprocapp import get_args
```
and then you will pass these variables to `get_args`

Did you try that and couldn't make it work which is why you are doing it this way?  I know I suggested to do it this way at some point, but the advantage of doing it the other way if we can is that we actually test the api of hte program that the user will be using, because that parser sets some defaults and so our tests will be aware of those.

sbillinge (commented):
this is looking better.  The comment about using the parser also applies to PR 46, but I suggest to get this merged in one of these PRs and then merge those changes into the other (choose the one that is closest to being merged maybe?)

src/diffpy/labpdfproc/tests/test_tools.py
yucongalicechen: Oh this makes sense! I updated and made it an error.

src/diffpy/labpdfproc/tests/test_tools.py
yucongalicechen: Got it. I just tried get_args() and I think it needs mud to be a required input. So whenever I call actual_parser = get_args(), it'll always cause an error. Any idea how I might fix this?

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -105,16 +104,33 @@ def test_load_user_metadata(inputs, expected):
...
+        "Please provide key-value pairs in the format key=value. "
+        "For more information, use `labpdfproc --help.`",
+    ),
+    (
+        [["beamline", "=", "28ID-2"]],
+        "Please provide key-value pairs in the format key=value. "
+        "For more information, use `labpdfproc --help.`",
+    ),
+    ([["facility=NSLS II", "facility=NSLS III"]], ["Please do not specify repeated keys: facility. "]),
+    ([["wavelength=2"]], ["Please do not specify repeated keys: wavelength. "]),
```
sbillinge, line 125: This one might be confusing because the user didn't specify repeated keys.  maybe `wavelength is a reserved name.  Please rerun using a different key name`?

We may want to store a list of reserved key names (it will basically  be `vars(args).keys()`).  We could list this in the args help if we want.

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -63,11 +63,10 @@ def get_args():
...
         nargs="+",
-        help="Specify (multiple) key-value pairs to be loaded into metadata using the format key=value. "
-        "Please separate each pair with whitespace, and make sure no whitespaces before or after the = sign. "
-        "Avoid using = in keys. If multiple = signs are present, only the first one separates the key and value. "
-        "If you repeat key names, it will replace previous values. "
-        "If a key or value contains whitespace, enclose it in quotes. "
+        help="Specify key-value pairs to be loaded into metadata using the format key=value. "
+        "Separate pairs with whitespace, and ensure no whitespaces before or after the = sign. "
+        "Avoid using = in keys. If multiple = signs are present, only the first separates the key and value. "
+        "Please do not repeat key names. If a key or value contains whitespace, enclose it in quotes. "
```
sbillinge, line 69: I think I would remove "please do not repeat key  names" from the help.  This is probably tmi.  I think as a user I eould only do this by mistake and the if I receive an error message saying "dumbass....you gave the same key twice" I would oh, I am a dumbass.    If we make the help too long people won't read and they won't see other, more important this.

Also, along the same lines, I think examples are better than text, so reproduce here some of your test examples of valid, well formed strings.  e.g., say, "for example, ....... are all valid key=value items"

sbillinge (commented):
The tests are getting very good now, congrats!  I like it very much how you are thinking.

We are nearly there, I just had a few comments.

sbillinge:
fantastic! nice work!

PR #49, opened 2024-05-06
changed output directory and wavelength tests to use get_args

yucongalicechen:


src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -1,78 +1,74 @@
...
+from diffpy.labpdfproc.labpdfprocapp import get_args
 from diffpy.labpdfproc.tools import known_sources, set_output_directory, set_wavelength
 
 params1 = [
-    ([None], ["."]),
-    (["."], ["."]),
-    (["new_dir"], ["new_dir"]),
-    (["existing_dir"], ["existing_dir"]),
+    ([], ["."]),
+    (["-o", "."], ["."]),
```
sbillinge, line 11: in tests I prefer that we always use the long name `--output_dir` or whatever it is, to make the test more readable.

sbillinge (commented):
this looks good .  Please can you see inline comments

src/diffpy/labpdfproc/tests/test_tools.py
yucongalicechen: I think the string here needs to be one of the arguments for user to specify? For example user specifies -o instead of output_directory. If we put output_directory it will cause an error saying it's not one of the arguments. Not sure if what I say here completely makes sense.

sbillinge:
We are after the full input with -- in front, not the abbreviated one with
a single dash

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -1,78 +1,77 @@
...
+    (["--output-directory", "input_dir"], ["input_dir"]),
 ]
 
 
 @pytest.mark.parametrize("inputs, expected", params1)
-def test_set_output_directory(inputs, expected, tmp_path):
-    directory = Path(tmp_path)
-    os.chdir(directory)
+def test_set_output_directory(inputs, expected, user_filesystem):
+    tmp_dir = user_filesystem
```
sbillinge, line 19: delete this line.

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -1,78 +1,77 @@
...
+    expected_output_directory = tmp_dir / expected[0]
```
sbillinge, line 20: change to something like `expected_output_directory = user_filesystem / expected[0]` (I am not sure if we need to use `Path(user_filesystem)` or sthg.... as I am not sure what the fixture returns.....bbut you can figure it out.  This works what you are doing, but again, I am trying to make the code more readable and understandable in the future

sbillinge (commented):
This looks great.  If you could just resolve the conflict and fix the one inline comment I can merge.

PR #50, opened 2024-05-07
load input directory and files

### LINKED ISSUE #23: specify an inputs directory name

yucongalicechen:
closes #23, #24

### DIFF
--- a/src/diffpy/labpdfproc/tests/test_tools.py
+++ b/src/diffpy/labpdfproc/tests/test_tools.py
@@ -17,7 +17,7 @@ from diffpy.utils.parsers.loaddata import loadData
 # Use cases can be found here: https://github.com/diffpy/diffpy.labpdfproc/issues/48
 
 # This test covers existing single input file, directory, a file list, and multiple files
-# We store absolute paths into input_directory and file names into input_file
+# We store absolute path into input_directory and file names into input_file
 params_input = [
     (["good_data.chi"], [".", "good_data.chi"]),  # single good file, same directory
     (["input_dir/good_data.chi"], ["input_dir", "good_data.chi"]),  # single good file, input directory
@@ -35,14 +35,14 @@ params_input = [
             ["good_data.chi", "good_data.xy", "good_data.txt", "unreadable_file.txt", "binary.pkl"],
         ],
     ),
-    (  # list of files provided (we skip if encountering an invalid files)
+    (  # list of files provided (we skip if encountering invalid files)
         ["good_data.chi", "good_data.xy", "unreadable_file.txt", "missing_file.txt"],
         [
             ".",
             ["good_data.chi", "good_data.xy", "unreadable_file.txt"],
         ],
     ),
-    (  # list of files provided (with invalid files and files in different directory)
+    (  # list of files provided (with invalid files and files in different directories)
         ["input_dir/good_data.chi", "good_data.xy", "missing_file.txt"],
         [
             ".",
@@ -53,7 +53,7 @@ params_input = [
         ["file_list_dir/file_list.txt"],
         [".", ["good_data.chi", "good_data.xy", "good_data.txt"]],
     ),
-    (  # file_list_example2.txt list of files provided with different paths
+    (  # file_list_example2.txt list of files provided in different directories
         ["file_list_dir/file_list_example2.txt"],
         [".", ["input_dir/good_data.chi", "good_data.xy", "input_dir/good_data.txt"]],
     ),
--- a/src/diffpy/labpdfproc/tests/test_tools.py
+++ b/src/diffpy/labpdfproc/tests/test_tools.py
@@ -16,34 +16,47 @@ from diffpy.utils.parsers.loaddata import loadData
 
 # Use cases can be found here: https://github.com/diffpy/diffpy.labpdfproc/issues/48
 
-# This test covers existing single input file, directory, or a file list
-# We store absolute path into input_directory and file names into input_file
+# This test covers existing single input file, directory, a file list, and multiple files
+# We store absolute paths into input_directory and file names into input_file
 params_input = [
-    (["good_data.chi"], [".", "good_data.chi"]),
-    (["input_dir/good_data.chi"], ["input_dir", "good_data.chi"]),
-    (["./input_dir/good_data.chi"], ["input_dir", "good_data.chi"]),
-    (
+    (["good_data.chi"], [".", "good_data.chi"]),  # single good file, same directory
+    (["input_dir/good_data.chi"], ["input_dir", "good_data.chi"]),  # single good file, input directory
+    (  # glob current directory
         ["."],
         [
             ".",
             ["good_data.chi", "good_data.xy", "good_data.txt", "unreadable_file.txt", "binary.pkl"],
         ],
     ),
-    (
+    (  # glob input directory
         ["./input_dir"],
         [
             "input_dir",
             ["good_data.chi", "good_data.xy", "good_data.txt", "unreadable_file.txt", "binary.pkl"],
         ],
     ),
-    (
-        ["input_dir"],
+    (  # list of files provided (we skip if encountering an invalid files)
+        ["good_data.chi", "good_data.xy", "unreadable_file.txt", "missing_file.txt"],
         [
-            "input_dir",
-            ["good_data.chi", "good_data.xy", "good_data.txt", "unreadable_file.txt", "binary.pkl"],
+            ".",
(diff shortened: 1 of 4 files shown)


sbillinge (commented):
please see my inline comment.  Also, please double check to have one test per UC and put a comment which UC it addresses.  Also in a comment above, copy-paste teh URL to the GH ussue where these are described
example of what I mean:
```
# Use cases can be found here: \https:github.com.......
["--input-file", "good_data.chi"], [".", "good_data.chi"]  # UC 1
```

I am only  looking at tests, not the code yet, but I am not clear what you are testing, in other words, for the most simple one of user specifies a single file....what is it that you want the code to do after that is specified?  This is what we want a test for.  Right now, I don't see that in these tests.  If you want to discuss it on slack, that is ok.  Don't bother working on the implementation code until we have written the tests.

sbillinge (commented):
please see my inline comment.

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -4,7 +4,80 @@
...
+def test_set_input_files(inputs, expected, user_filesystem):
+    expected_input_directory = Path(user_filesystem) / expected[0]
+    expected_input_files = expected[1]
+
+    cli_inputs = ["2.5"] + inputs
+    actual_args = get_args(cli_inputs)
+    actual_args = set_input_files(actual_args)
+    assert actual_args.input_directory == expected_input_directory
+    assert set(actual_args.input_file) == set(expected_input_files)
+
```
sbillinge, line 55: These are good tests.

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -4,7 +4,80 @@
...
+
+params_input_bad = [
+    (["new_file.xy"]),
```
sbillinge, line 58: `non_existing_file.xy`?

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -4,7 +4,80 @@
...
+    (["./input_dir/new_file.xy"]),
```
sbillinge, line 59: see above?

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -4,7 +4,80 @@
...
+    (["./new_dir"]),
```
sbillinge, line 60: see above comment

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -4,7 +4,80 @@
...
+]
+
+
+@pytest.mark.parametrize("inputs", params_input_bad)
+def test_set_input_files_bad(inputs, user_filesystem):
+    cli_inputs = ["2.5"] + inputs
+    actual_args = get_args(cli_inputs)
+    with pytest.raises(ValueError):
```
sbillinge, line 68: Is this your desired behavior.  Is the error message satisfactory to help the user debug (see (4.) above)

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -4,7 +4,80 @@
...
+        actual_args = set_input_files(actual_args)
+
+
+def test_loadData_with_input_files(user_filesystem):
```
sbillinge, line 72: This seems to be testing your test, not anything in the code?  I guess this is ok, but maybe make a comment to that effect?

sbillinge (commented):
The tests look ok, please see inline comments.

I don't see tests for 1., 2., 3., 5., 6.  in my list above. Some of these will be covered by other functions, but it is good to comment on what and how.

Also, I don't see a similar list for the list and glob situations.

src/diffpy/labpdfproc/tools.py
```diff
@@ -1,9 +1,40 @@
...
+    Returns
+    -------
+    args argparse.Namespace
+
+    """
+    if not Path(args.input).exists():
+        raise ValueError("Please specify valid input file or directory.")
+
+    if not Path(args.input).is_dir():
+        input_dir = Path.cwd() / Path(args.input).parent
```
sbillinge, line 27: I wonder whether this will fail if the user gives a filename with a path that doesn't include cwd?  For example, I think a valid test would be:
```
cd /user/me/analysis
labpdfcor 2.5 /user/me/data/my_file.xy
```
Please could you add test for this situation and make sure it passes?

src/diffpy/labpdfproc/tools.py
```diff
@@ -1,9 +1,40 @@
...
+        input_file_name = Path(args.input).name
+    else:
+        input_dir = Path(args.input).resolve()
+        input_files = [file for file in glob.glob(str(input_dir) + "/*", recursive=True) if os.path.isfile(file)]
+        input_file_name = [os.path.basename(input_file_path) for input_file_path in input_files]
+    setattr(args, "input_directory", input_dir)
+    setattr(args, "input_file", input_file_name)
+    return args
```
sbillinge, line 35: otherwise, I think this looks good.

sbillinge (commented):
OK, progress.  Please see inline comments.

However, I don't see where you handle the lists of files.  For example, if there is a list of files we presumably need to apply the correction to each file in the list and then write into the header of that file its parent file, not a glob list.

yucongalicechen:
I added tests for file list and edited help message for it. Currently, the program reads the file as a file list only if every line in it is a valid file name within the same directory of the file list.

sbillinge (commented):
Here is a slight tweak to the tests, and then I think we can move to close.

```
# Use cases can be found here: https://github.com/diffpy/diffpy.labpdfproc/issues/48

params_input = [
    (["good_data.chi"], [".", ["good_data.chi"]]),   # single good file, same directory
    (["input_dir/good_data.chi"], ["input_dir", ["good_data.chi"]]),  # single good file, input directory
    (  # glob current directory
        ["."],
        [
            ".",
            ["good_data.chi", "good_data.xy", "good_data.txt", "unreadable_file.txt", "binary.pkl"],
        ],
    ),
    (  # glob input directory
        ["./input_dir"],
        [
            "input_dir",
            ["good_data.chi", "good_data.xy", "good_data.txt", "unreadable_file.txt", "binary.pkl"],
        ],
    ),
    (  # list of files provided
        ["input_dir/good_data.chi ./good_data.chi unreadable_file.txt missing_file.txt"], ["input_dir", ["good_data.chi", "good_data.xy", "good_data.txt"]]),
    (  # file_list.txt list of files provided
        ["file_list_dir/file_list.txt"], ["file_list_dir", ["./good_data.chi", "input_dir/good_data.chi",  "./good_data.xy"]]),
]
```

sbillinge:
btw, for lists and globs I think the best behavior would be to read as many files as possible and skip over the others, letting the user know which have been skipped.

Later when we make a function to do the work and we write tests for it, let's have tests for all the same sets of inputs as here and check that the right files are written to the right places (so glob the directories after the corrections are applied and check the list is as expected) and check the printed outputs are correct for the skipped files.  Also, let's make sure that the args that are written to the headers are correct, where the parent file is correctly identified (i.e., not just the input file list saved to the header, but also the actual file that was used for that particular correction.  This will have to be set inside the loop).

Please maybe copy-paste these comments to a new issue for that function.

yucongalicechen:
I added tests for testing file lists and multiple files. Don't know if this is the best for distinguishing between a file list and a data file: currently I read the first line of a file and if it is not a existing file name, I treat it as a data file.

src/diffpy/labpdfproc/tools.py
```diff
@@ -1,9 +1,77 @@
...
+    Otherwise if we have a directory, glob all files within it.
+    If there are any invalid filenames (for the cases of multiple files, file list, or directory), we skip them.
+
+    Returns
+    -------
+    args argparse.Namespace
+
+    """
+
+    if len(args.input) > 1:
```
sbillinge, line 31: force the input to be a list and remove this.

sbillinge (commented):
Please see my inline comments.

On balance the function looks a bit complicated.  Let's make sure there is not a simpler way of doing it.

yucongalicechen:
I edited help message, args.input is now a list, edited the test for same file names in different directories (if this happens, we probably would want to change output name/directory too, since they are the same name in the same directory?). Now it seems that we only need input_directory to get absolute paths for each file (i.e. input_file_name is unncessary, I removed it).
I simplied the input function a bit: for every input, we check if it is directory/datafile/file list, use try/except to raise an error message if any file is invalid, and raise an error if all files are invalid.

PR #58, opened 2024-05-13
closes #29: initial commit on  load datetime of the analysis

### LINKED ISSUE #29: load the datetime of the analysis into metadata

stevenhua0320:
Initial commit for the load_datetime. Not sure about whether it is automatically extracted or manually created. So, in the test function, I write the version of manually written ones. If it is the former, let me change it to the correct form. Should take some return objects after executing the function. More modification needs to be done in the 2nd push of the function.

### DIFF
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -89,6 +89,15 @@ def get_args(override_cli_inputs=None):
         "For example, facility='NSLS II', 'facility=NSLS II', beamline=28ID-2, "
         "'beamline'='28ID-2', 'favorite color'=blue, are all valid key=value items. ",
     )
+    p.add_argument(
+        "-d"
+        "--datetime_of_analysis",
+        metavar="YYYY-MM-DD HH:MI:SS",
+        help=f"The date and time of the analysis. Required."
+             "The format should be in YYYY-MM-DD HH:MM:SS format.",
+    )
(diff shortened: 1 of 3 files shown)


yucongalicechen:
Good start! A few comments:

1. It is better if we don’t require user to enter date time manually, and instead fetch the time automatically. There are a few functions that can do this.
2. It might be better to call the args in tests actual_args and expected_args because these args include more than datetime argument.
3. In the tests, we need to give inputs for both mud and input (see wavelength tests). I believe you also need to do args = load_datetime in order to load the error message.

Please try to include more information in the commit message, maybe something like “initial commit on requiring users to manually enter a datetime”.

sbillinge:
Here is a suggested UC
UC 1 user searching for files
==================

1. chem grad student (CGS) wants to search for files that they worked on on Thursday afternoon.
2. CGS writes a quick script in jupyter that searches over files in a directory tree, obtains the datetime from the header and sorts appropriately, returning a list of files that pass the filter

In order for this UC to work we need another
UC2 labpdfproc writing datetime to header
============================

1. User runs labpdfproc (LPP)
2. LPP successfully executes to compute a correction for each file
3. LPP adds the datetime to metadata before writing the output file
4. LPP adds the metadata to the file header

For testing we will need to have a fake datetime.  We can do this by mocking datetime.  The code will need to know whether it is a test running or not.  There seem to be some hacky ways to do it but it may be better to be explicit and allow a `--is-test` CLI input that is false by default but can be set to true in all our tests.

Issue #52, opened 2024-05-09
Implement input function to handle skipped files and check ouputs and header


yucongalicechen:
For lists and globs, the best behavior would be to read as many files as possible and skip over the others, letting the user know which have been skipped.
We would like to have tests for all the same sets of inputs and check that the right files are written to the right places (so glob the directories after the corrections are applied and check the list is as expected) and check the printed outputs are correct for the skipped files. Also, let's make sure that the args that are written to the headers are correct, where the parent file is correctly identified (i.e., not just the input file list saved to the header, but also the actual file that was used for that particular correction. This will have to be set inside the loop).

PR #59, opened 2024-05-15
implement wildcard pattern for input

yucongalicechen:
- Initial commit
- Related to Issue #52, examples are here #48
- I skipped file_list files as usual. Shall we allow the user to specify something like "file*"?

### DIFF
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -3,7 +3,6 @@ from argparse import ArgumentParser
 
 from diffpy.labpdfproc.functions import apply_corr, compute_cve
 from diffpy.labpdfproc.tools import (
-    expand_list_file,
     known_sources,
     load_user_metadata,
     set_input_lists,
@@ -29,14 +28,9 @@ def get_args(override_cli_inputs=None):
         "'.' (load everything in the current directory), 'data' (load"
         "everything in the folder ./data), 'data/file_list.txt' (load"
         " the list of files contained in the text-file called "
-        "file_list.txt that can be found in the folder ./data). "
-        "\nWildcard character (*) is accepted. Examples include './*.chi'"
-        " (load all files with .chi extension), 'data/*.chi' (load all "
-        "files in 'data' file with .chi extension), 'file*.chi' (load all "
-        "files starting with 'file' and ending with .chi extension), 'test*' "
-        "(load all files and directories starting with 'test'), 'test*/*.chi' "
-        "(load all directories starting with 'test' and all files under "
-        "with .chi extension). ",
+        "file_list.txt that can be found in the folder ./data), "
+        "'data/*.chi' (load all files with extension .chi in the "
+        "folder ./data), 'data*' (load all files and directories starting with 'data').",
     )
     p.add_argument(
         "-a",
@@ -102,7 +96,6 @@ def get_args(override_cli_inputs=None):
 
 def main():
     args = get_args()
-    args = expand_list_file(args)
     args = set_input_lists(args)
     args.output_directory = set_output_directory(args)
     args.wavelength = set_wavelength(args)
@@ -126,7 +119,7 @@ def main():
             )
 
         input_pattern = Diffraction_object(wavelength=args.wavelength)
-        xarray, yarray = loadData(args.input_file, unpack=True)
+        xarray, yarray = loadData(filepath, unpack=True)
         input_pattern.insert_scattering_quantity(
             xarray,
             yarray,
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -29,8 +29,8 @@ def get_args(override_cli_inputs=None):
         "everything in the folder ./data), 'data/file_list.txt' (load"
         " the list of files contained in the text-file called "
         "file_list.txt that can be found in the folder ./data), "
-        "'data/*.chi' (load all files with extension .chi in the "
-        "folder ./data), 'data*' (load all files and directories starting with 'data').",
+        "'./*.chi', 'data/*.chi' (load all files with extension .chi in the "
+        "folder ./data).",
     )
     p.add_argument(
         "-a",
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -30,9 +30,13 @@ def get_args(override_cli_inputs=None):
         "everything in the folder ./data), 'data/file_list.txt' (load"
         " the list of files contained in the text-file called "
         "file_list.txt that can be found in the folder ./data). "
-        "Wildcard character (*) is accepted. Examples include './*chi'"
-        " (load all files with .chi extension) and 'data/test*' (load "
-        "all files starting with 'test' in the folder ./data). ",
+        "\nWildcard character (*) is accepted. Examples include './*.chi'"
+        " (load all files with .chi extension), 'data/*.chi' (load all "
+        "files in 'data' file with .chi extension), 'file*.chi' (load all "
+        "files starting with 'file' and ending with .chi extension), 'test*' "
+        "(load all files and directories starting with 'test'), 'test*/*.chi' "
+        "(load all directories starting with 'test' and all files under "
+        "with .chi extension). ",
     )
     p.add_argument(
         "-a",
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -29,7 +29,10 @@ def get_args(override_cli_inputs=None):
         "'.' (load everything in the current directory), 'data' (load"
         "everything in the folder ./data), 'data/file_list.txt' (load"
         " the list of files contained in the text-file called "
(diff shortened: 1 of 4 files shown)


sbillinge (commented):
very nicely done.  Just one small typo and one suggestion in the input.

I also see some copy-pasted code which makes me wonder if it might be better to simply look for the wildcard and expand it before passing into this code block.  Kind of like a prefilter that does something like
```
for input_name in args.input:
     if '*" in name 
             do the globbing and find the list of files
             add the glob list to args.inputs and remove the "*" name
```

then again,
```
for input_name in args.inputs:
    code unchanged from before
```
which would be easier to maintain.

yucongalicechen:
- I added a function expand_wildcard_file to take care of wildcard patterns and return the expanded list of inputs. Shall we put the two expand functions into set_input_lists for convenience? So that in the labpdfproc.py and tests, we only need to call for set_input_lists.
- I wrote an error message for wildcard patterns that cannot match any files. But if the user specifies an invalid wildcard pattern like "**", this error message might not be that helpful as the original error message from Python (as it would tell user that it is an invalid wildcard pattern). I currently cannot think of a way to distinguish between these two situations, unless checking that if wildcard pattern is one of the invalid cases, but there are many invalid cases to check..

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -29,7 +29,14 @@ def get_args(override_cli_inputs=None):
         "'.' (load everything in the current directory), 'data' (load"
         "everything in the folder ./data), 'data/file_list.txt' (load"
         " the list of files contained in the text-file called "
-        "file_list.txt that can be found in the folder ./data).",
+        "file_list.txt that can be found in the folder ./data). "
+        "\nWildcard character (*) is accepted. Examples include './*.chi'"
+        " (load all files with .chi extension), 'data/*.chi' (load all "
+        "files in 'data' file with .chi extension), 'file*.chi' (load all "
```
sbillinge, line 35: This is probably a bit too long now.  I htink most people know how "*" wildcards work, so i would just suggest to use the `data/*.chi` example and one example for a directory, maybe `inputs*` which in principle would open all files in any directory that begins with "inputs" but also any file that begins with "inputs" in the current directory.

We may want to think about not letting the users specify directories with "*".  It could cause problems and also not be that useful.  For example, if they just specify "*" it will glob all subdirectories of their current positoin and operate on all files in all the directories.  if they do this from a home directory or something it could give unwanted behavior. 

On balance, I think this might be better, so the wildcard only applies to files in the specified directory.  Then if they want to do this over multiple directories they need to specify this as a list, and if it is a long list it can be done in an `input-files.txt` file.

src/diffpy/labpdfproc/tools.py
```diff
@@ -28,6 +29,41 @@ def set_output_directory(args):
...
+    wildcard_inputs = [input_name for input_name in args.input if "*" in input_name]
+    for wildcard_input in wildcard_inputs:
+        if not glob.glob(wildcard_input):
+            raise FileNotFoundError(
+                f"Invalid wildcard input {wildcard_input}. "
+                f"Please ensure the wildcard pattern matches at least one file or directory."
+            )
+        input_files = Path(".").glob(wildcard_input)
+        for input_file in input_files:
+            if input_file.is_file():
```
sbillinge, line 55: I think some of the logic here is handled later.  I would just operate on strings here.
```
for wildcard_input in wildcard_inputs:
    input_files = Path(".").glob(wildcard_input)
    args.input.extend(input_files)
    args.input.remove(wildcard_input)
```
and that's it?

sbillinge (commented):
please see inline comments.

sbillinge:
re your comment, I agree that it might be good to put the two pre-filters into the `expand_user_input` function. Perhaps we shoud also make them private functions (add an underscore at the beginning) which makes the code more readable.  It is nice that we have independent tests for them, no need to undo that.

src/diffpy/labpdfproc/labpdfprocapp.py
yucongalicechen: I agree. I simplified and combined the expand functions.

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -29,7 +28,9 @@ def get_args(override_cli_inputs=None):
         "'.' (load everything in the current directory), 'data' (load"
         "everything in the folder ./data), 'data/file_list.txt' (load"
         " the list of files contained in the text-file called "
-        "file_list.txt that can be found in the folder ./data).",
+        "file_list.txt that can be found in the folder ./data), "
+        "'data/*.chi' (load all files with extension .chi in the "
+        "folder ./data), 'data*' (load all files and directories starting with 'data').",
```
sbillinge, line 33: we are not now supporting this so please remove this one.

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -54,6 +53,32 @@
...
+    ),
+    (  # wildcard pattern, matching files with .chi extension in the input directory
+        ["input_dir/*.chi"],
+        ["input_dir/good_data.chi"],
+    ),
+    (  # wildcard pattern, matching files starting with good_data
+        ["good_data*"],
+        ["good_data.chi", "good_data.xy", "good_data.txt"],
+    ),
+    (  # wildcard pattern, matching files or directories starting with input
```
sbillinge, line 68: we are not supporting this, so this can be removed.

src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -54,6 +53,32 @@
...
+        ["input*"],
+        [
+            "input_dir/good_data.chi",
+            "input_dir/good_data.xy",
+            "input_dir/good_data.txt",
+            "input_dir/unreadable_file.txt",
+            "input_dir/binary.pkl",
+        ],
+    ),
+    (  # wildcard pattern, matching files or directories starting with unreadable and ending with .txt extension
```
sbillinge, line 78: remove.

src/diffpy/labpdfproc/tools.py
```diff
@@ -42,7 +42,14 @@ def expand_list_file(args):
     the arguments with the modified input list
 
     """
+    wildcard_inputs = [input_name for input_name in args.input if "*" in input_name]
     file_list_inputs = [input_name for input_name in args.input if "file_list" in input_name]
+
```
sbillinge, line 47: close up blank line

src/diffpy/labpdfproc/tools.py
```diff
@@ -70,6 +77,7 @@ def set_input_lists(args):
     """
 
     input_paths = []
+    args = _expand_user_input(args)
```
sbillinge, line 80: this is very pretty code now....I like it!

sbillinge (commented):
we may need a test for a user putting a wildcard in the file-list file.  I think your code may currently fail this test, but let's make the test and see.

Almost there, see a few last cleaning edits, and also I think my request to not support wildcards on directories may have gotten lost in those long comments on the last review, bu tplease remove support for wildcards on directories.

src/diffpy/labpdfproc/tools.py
yucongalicechen: I added the test case of having a wildcard pattern in a file-list file. Also, I want to make sure that in the wildcard pattern, we also don't want to read anything related to a file-list file, to avoid confusion? In the test case I just added it is to glob all txt files, that include a file-list file, which I skipped on purpose.

src/diffpy/labpdfproc/tools.py
```diff
@@ -48,6 +48,11 @@ def expand_list_file(args):
             file_inputs = [input_name.strip() for input_name in f.readlines()]
         args.input.extend(file_inputs)
         args.input.remove(file_list_input)
+    wildcard_inputs = [input_name for input_name in args.input if "*" in input_name]
+    for wildcard_input in wildcard_inputs:
+        input_files = [str(file) for file in Path(".").glob(wildcard_input) if "file_list" not in file.name]
```
sbillinge, line 53: I think we may be able to remove `if file_list not in file.name` because file_list files have been removed already.

src/diffpy/labpdfproc/tests/conftest.py
```diff
@@ -36,6 +36,8 @@ def user_filesystem(tmp_path):
     with open(input_dir / "binary.pkl", "wb") as f:
         f.write(binary_data)
 
+    with open(base_dir / "file_list.txt", "w") as f:
+        f.write("./*.txt")
     with open(input_dir / "file_list.txt", "w") as f:
```
sbillinge, line 41: This is fine, though we could have tested it with a single file_list.txt file with the right contents.

sbillinge (commented):
nice work.   I could merge it like this, but there are a couple of final comments.

src/diffpy/labpdfproc/tools.py
yucongalicechen: I think this is for files in the glob directory (not in args.input), so if we have a file list in the same directory as the wildcard, then it'll be loaded if we don't skip it.

sbillinge: I see, yes better to have it to be on the safe side.   It may be better to make it stricter so `if file.name == file_list.txt` but it is probably ok as it is.

PR #62, opened 2024-05-17
load username, email, and orcid number into config files

### LINKED ISSUE #30: load the username into the metadata
load the username into the metadata (try and get it from system but allow user to overload that?)

### LINKED ISSUE #31: load user email into metadata

stevenhua0320:
Closes #30
Closes #31

### DIFF
--- a/src/diffpy/labpdfproc/tests/test_tools.py
+++ b/src/diffpy/labpdfproc/tests/test_tools.py
@@ -8,6 +8,7 @@ from diffpy.labpdfproc.labpdfprocapp import get_args
 from diffpy.labpdfproc.tools import (
     known_sources,
     load_user_metadata,
+    load_username_email,
     set_input_lists,
     set_output_directory,
     set_wavelength,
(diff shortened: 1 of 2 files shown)


sbillinge (commented):
Great start, thanks Steven.  You did a really good job here!  Here are some comments.

1. please see how I edited the PR title.  Your commit message is good but do not use a commit measage for a PR title, they do a different job.
2. for the issues that this closes (based on the title) I think there will be two or three issues, not just one  you will need a closes `#` for each one.
3. per our discussion, there will not be a separate function for this task but it will be handled in the function that handles the flow.  this is a difficult PR to atm because we didn't resolve that conversation, but I can still give feedback on what you did, so that is great!
4. you have three tests but there is some redundancy in that multiple tests test the same thing (user enters username and email but not orcid at the command line)
5. .please see inline comments.
6. for this problem the more interesting tests will be the ones that don't handle entry on the command line.  they will be a bit harder to write too.  

But this is a good start.  As I said, this is a slightly harder PR, so please maybe open the other ones you are working as well and I can give feedback on those too.

Issue #56, opened 2024-05-11
Documents and build


sbillinge:
use diffpy.utils as a template

sbillinge:
closed as completed

PR #66, opened 2024-05-20
GH action to build and deploy docs

cadenmyers13:
closes [#56](https://github.com/diffpy/diffpy.labpdfproc/issues/56) 

Most of this code is from `docs.yml` in `diffpy.utils`


sbillinge:
@cadenmyers13 this is failing flake8 in pre-commit.  You can see what the failure is by clicking on "Details".  You will have to fix those two issues identified manually and then try again.  If you merge a synced main into this branch then you will get the pre-commit working on your local and you can check this passes before pushing to the PR.

You can also see why the build docs is failing by clicking on details in that workflow.  We are missing a build.txt with the requirements needed for building the docs.

.github/workflows/docs.yml
```diff
@@ -37,6 +41,7 @@ jobs:
         run: make -C doc/manual html
 
       - name: Deploy
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
```
Sparks29032, line 44: if: ${{ (github.event_name == 'push' && github.ref == 'refs/heads/main') || github.event_name == 'release' }}

PR #71, opened 2024-05-23
GH action to build and deploy docs

### LINKED ISSUE #56: Documents and build

cadenmyers13:
closes #56 

Created a new PR because I was pushing from main on the other one. 

This should fix the flake8 pre-commit issue


sbillinge (commented):
Thanks @cadenmyers13 .   I think we are not doing the `pull_request:` and the related `if` in `doc.yml`

Also, I am worried that your main is not clean as I keep seeing a commit in your PRs which is a merge of main.   Also there are multiple things on this PR.   

I could merge it like this (when you fix the `pull_request:`), but it would be good to make sure everything is ok with your setup and your workflows, so I suggest that you make a new PR on a clean branch that just does the GH action and nothing else.

cadenmyers13:
> Thanks @cadenmyers13 . I think we are not doing the `pull_request:` and the related `if` in `doc.yml`
> 
> Also, I am worried that your main is not clean as I keep seeing a commit in your PRs which is a merge of main. Also there are multiple things on this PR.
> 
> I could merge it like this (when you fix the `pull_request:`), but it would be good to make sure everything is ok with your setup and your workflows, so I suggest that you make a new PR on a clean branch that just does the GH action and nothing else.

Yeah I've kind of made a mess with my branches. After this PR, i will sort everything out

sbillinge:
closing PR so we can sort out branches and do it cleanly.

PR #72, opened 2024-05-23
implement functionality to allow input selection or skipping of files

yucongalicechen:
Addressing issue #52

I went through the UCs and test cases in #48 and there left these cases (I addressed UC7 and -file7,chi in this PR and will probably address UC5 in another PR):
1. UC5: multiple files, different samples (file1.chi 2.5 file2.chi 3.5)
2. UC7: automatically selects the right files to apply the correction to (don't read _corrected and _cve).
3. -file7.chi (specify files don't want to read).

For 2, I didn't add any additional test cases and put some corrected files in our tmp path.
For 3, since - would be taken as one of the arguments, I changed it to ~.


sbillinge:
I don't think we are implementing these UCs in this release.  That are
something someone might want to do but we will wait to see how users use
the code before implementing them

PR #74, opened 2024-05-30
Load datetime of analysis, package name and version into metadata

### LINKED ISSUE #28: load the name of the package and the version number into metadata

stevenhua0320:
Closes #28 #29 
Initial commit to dynamically load the datetime of analysis, package name, and package version into metadata using monkeypatch. Not sure whether it should be set as an attribute or directly loaded into it.


sbillinge (commented):
I think we decided that this functionality was going to go into the diffraction_objects, didn't we @yucongalicechen ?

yucongalicechen:
Yeah the function is added to apply_corr() in the corrected diffraction objects. Unless you meant to add it to diffpy.utils?

sbillinge:
> Yeah the function is added to apply_corr() in the corrected diffraction objects. Unless you meant to add it to diffpy.utils?

We probably want it in the diffraction_objects themselves. but this will capture the name and version of diffpy.utils used for the diffraction object.  If we also want to capture the version of labpdfproc then we will also need it in apply_corr().  

How about this for a design:

1.  a public function in diffpy.utils that captures package name and version number and  formats this for dumping
3.  functionality in diffraction_object dump function that calls this function and loads it into metadata at dump time
4. call this function in apply_corr() and load it into metadata.

We need to think what we want the dumped file header to look like that is the most useful.  Do you guys want to suggest something?

sbillinge:
e.g., in the file, a nice human readable thing might be:
```
package_versions: diffpy.labpdfproc v1.3.4, diffpy.utils v2.4.5
```
but it could be something different.

stevenhua0320:
> e.g., in the file, a nice human readable thing might be:
> 
> ```
> package_versions: diffpy.labpdfproc v1.3.4, diffpy.utils v2.4.5
> ```
> 
> but it could be something different.

Yeah, I think these information is enough for users to obtain useful information about the package number and version for the file. So, does that mean that we also need to define the functionality of load package name and version in diffpy.utils, probably implement with a public function?

yucongalicechen:
- Just want to make sure that for package versions, we want to load both labpdfproc and diffpyutils packages instead of overriding, so we're adding labpdfproc version when we call the function there?
- Would it be better if we pass the package version as an argument to dump like what was done with xtype=None? So that we can take care of package version outside of dump and just write what's needed.
- We can add the time attribute directly in dump function?

sbillinge:
I can get started now:

Uc1

1. Diffraction object wants to save it's version on s human readable form
in an output file by default
2. Diffpy.utils does this by saving it's version

Uc2
1. Labpdfproc wants to save it's version on s human readable form in an
output file by default
2. LPP is using diffraction objects so it does it by extending the behavior
in DO

Uc3
1. As 1.2 but for pdfgetx3 but X3 is refactored to use DO

Uc4
1. Diffraction programmer is writing their own program and not using DO and
want to use their own formatting but have diffpy.utils already as a
dependency
2. They use a helper function from utilities to easily get the information
then use it to write their own output code

sbillinge:
Ok, that's it

sbillinge:
Uc3 and uc2 are the same UC but this highlights the level of reusability we
are looking for

yucongalicechen:
Got it, I understand now. We'll start working on it soon!

stevenhua0320:
> @stevenhua0320 I am closing this. Please could you update your diffpy.utils (you will have to install it in develop mode as we haven't released it to conda-forge yet, @yucongalicechen can help), then make a new branch on a clean main and redo this making use of the `package_info` function in `diffpy.utils.tools`
> 
> @yucongalicechen we probably need a PR with documentation in diffpy.utils for how to use that function. It might be good if you write that documentation first and we can see if Steven can follow it.....

OK, I would follow yucong's instructions on updating utils and then using utils functionality to load these infos in labpdfproc.

PR #77, opened 2024-06-11
loading username and email using function in diffpyutils

### LINKED ISSUE #30: load the username into the metadata

### LINKED ISSUE #31: load user email into metadata

yucongalicechen:
- closes #30, closes #31
- Edits in conftest.py and test_tools.py are basically the same as what's in diffpyutils, except for the private test function _run_tests I replaced the config dictionary with the args
- In labpdfprocapp.py: Added username and email arguments to allow cli inputs
- In tools.py: import the function from diffpyutils


src/diffpy/labpdfproc/tests/test_tools.py
```diff
@@ -241,3 +243,119 @@ def test_load_user_metadata_bad(inputs, msg):
     actual_args = get_args(cli_inputs)
     with pytest.raises(ValueError, match=msg[0]):
         actual_args = load_user_metadata(actual_args)
+
+
+def _setup_dirs(monkeypatch, user_filesystem):
```
sbillinge, line 248: I think most of these tests are not needed since they are tested in the `diffpy.utils.tools` functions.

We just have to test what this load_user_data does, so give it an arg without the username and have it return an arg with the username.  I guess we could test when the `arg.username` is `None` and when it has something in it, then just create a config file with something in it.  You don't have to test all the different config possibilities which are tested in utils.

sbillinge (commented):
load_user_input is nicely done.  there are way too many tests though, please see inline comment.

src/diffpy/labpdfproc/tests/test_tools.py
yucongalicechen: Ahh got it. In this case we only need to test args with/without username/email and check that arg.username and email are properly loaded?

sbillinge (commented):
nicely done!

This should pass when the diffpy.utils release appears on conda-forge.  Let's leave this PR open until it passes.  But no more work to do on this one.

yucongalicechen:
> nicely done!
> 
> This should pass when the diffpy.utils release appears on conda-forge. Let's leave this PR open until it passes. But no more work to do on this one.

Do the help messages need editing?

sbillinge:
Yes, I can do that

PR #78, opened 2024-06-14
test for compute_cve()

### LINKED ISSUE #20: test for `compute_cve()`

yucongalicechen:
- closes #20
- Initial commit
- I have to add the line `n_points_on_diameter=N_POINTS_ON_DIAMETER` to make monkeypatch work in the test. It doesn't affect the current functionality but is probably not the desired behavior.. Any suggestions for this?

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -43,7 +43,7 @@ jobs:
           conda config --add channels conda-forge
           conda activate test
           conda install --file requirements/requirements.txt
-          conda install pytest coverage codecov
+          conda install --file requirements/requirements-dev.txt
           pip install .
 
       - name: Validate diffpy.labpdfproc
--- a/requirements/requirements-dev.txt
+++ b/requirements/requirements-dev.txt
@@ -9,6 +9,7 @@ nbstripout
 pre-commit
 pre-commit-hooks
 pytest
+pytest-mock
 sphinx
 twine
 # These are dependencies of various sphinx extensions for documentation.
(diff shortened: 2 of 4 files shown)


sbillinge:
I would like to unify our mocking to pytest mocker.  Please search in
recent diffpy .I told where I did this

sbillinge:
this is failing CI.  It can't find mocker for some reason.  I don't immediately see the problem, but please take a look.

yucongalicechen:
Oh okay, I will look into this.
I'm also thinking if we want to update package info in this function. It might be easier to maintain/check if we load the package info into "input_pattern" DO along with other things in args. But would it make more sense to update the package here?

sbillinge:
> Oh okay, I will look into this. I'm also thinking if we want to update package info in this function. It might be easier to maintain/check if we load the package info into "input_pattern" DO along with other things in args. But would it make more sense to update the package here?

I'm not really sure.  I think in general we may want to update package_info whenever a package is used.  The way we implemented it was that it was a set, so each package would only appear once, regardless how often it is used.  From this point of view, we may want to just update labpdfproc once near the beginning of when it is used.  If it doesn't already, we may want diffpy_utils to update whenever a DO is instantiated, but that would happen in `diffpy.utils` not here.

tl;dr, I don't think we need it here, but I do think we need it in labpdfprocapp somewhere.

yucongalicechen:
Got it. So this would be similar to loading user info into args. I'll talk to Steven to see if he could work on that.

sbillinge:
nice!

PR #79, opened 2024-06-20
load package info

### LINKED ISSUE #28: load the name of the package and the version number into metadata

yucongalicechen:
closes #28


sbillinge (commented):
there may be a good reason, but why is there a merge of main into this branch?

yucongalicechen:
> there may be a good reason, but why is there a merge of main into this branch?

I merged because I wasn't super sure if it could find the moker I updated previously.. maybe not necessary?

sbillinge:
was it failing tests because it couldn't find that?  that would be the way to tell.

It is ok, I can merge it like this, but please practice good git hygiene.....don't merge main into branch unless absolutely needed.

PR #80, opened 2024-06-20
tests for apply_corr

### LINKED ISSUE #19: test for `apply_corr()` in `functions.py`

yucongalicechen:
closes #19
@sbillinge I saw that you assigned this to yourself but I thought I could just do it conveniently by following the tests for compute_cve

### DIFF
--- a/src/diffpy/labpdfproc/tests/test_functions.py
+++ b/src/diffpy/labpdfproc/tests/test_functions.py
@@ -1,7 +1,7 @@
 import numpy as np
 import pytest
 
-from diffpy.labpdfproc.functions import Gridded_circle, compute_cve
+from diffpy.labpdfproc.functions import Gridded_circle, apply_corr, compute_cve
 from diffpy.utils.scattering_objects.diffraction_objects import Diffraction_object
 
 params1 = [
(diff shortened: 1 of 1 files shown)


src/diffpy/labpdfproc/tests/test_functions.py
```diff
@@ -80,3 +80,39 @@ def test_compute_cve(mocker):
         scat_quantity="cve",
     )
     assert actual_abdo == expected_abdo
+
+
+def test_apply_corr(mocker):
```
sbillinge, line 85: this test is probably correct but hard to follow. 
1. there is a lot of duplicated code instantiating the DO's so maybe make a helper function called something like `_instantiate_test_do(yarray)` and just call that with only the y-array as input
2. maybe make the inputs easier to understand, so input y-array is [2., 2., 2.] and correction is [0.5, 0.5, 0.5] and so expected is a do with yarray [1., 1., 1.]. 

Would these make it easier to read?

If you make that wrapper function, on a separate PR we can maybe clean up some of the other tests as well.

sbillinge (commented):
please see inline

src/diffpy/labpdfproc/tests/test_functions.py
yucongalicechen: Sounds good, I'll do that

yucongalicechen:
closed by #81

Issue #85, opened 2024-07-23
Implement quick calculation


yucongalicechen:
(I can split the issues later if that's easier)
- In a "data" folder: (1) store precomputed data 1/cve for muD=1 (as an .xy file), (2) store coefficient list in a csv file.
- In a new module "fast_cve":
- (1) Define global variables: angles from 1-180 degrees, sparse dataset muD list and coefficient list, an interpolation function based on these coefficients.
- (2) A function (and its test function) that reads in the muD and uses the interpolation function to predict the cve curve, and interpolates it onto user's tth (similar to compute_cve in functions.py).
- (3) A function (and its test function) similar to apply_corr that returns the corrected intensities.
- In labpdfprocapp.py: change default behavior to using fast_cve but give an optional command -b to enable brute-force calculation (raise an error if muD < 0.5 or muD > 6?).

sbillinge:
looks good.  let's call the module `fast_cve` maybe?

Default behavior should be to use fast_cve but command-line override to do brute-force calculation

Write tests with the functions, so not on a separate activity.

yucongalicechen:
Sounds good! I edited the comment.

yucongalicechen:
updated by issue #91, closed by #92.

PR #86, opened 2024-07-24
implement fast_cve module

yucongalicechen:
- addressing issue #85 (will address the last bullet point about labpdfprocapp.py in another PR)
- functions and tests are very similar to what's in functions.py
- would like some advice on renaming some variables/functions to make them more readable


sbillinge:
wow, super, merged in one!!!!!!

PR #87, opened 2024-07-25
add fast_cve to labpdfprocapp.py and some other fixes

### LINKED ISSUE #85: Implement quick calculation

yucongalicechen:
- closes #85 (changed default calculation to fast cve while giving an option of -b to enable brute-force calculation. system exit if mu*D is out of range for 0.5-6.)
- fixed a few other issues: it seems that we accidentally deleted the labpdfproc command last time (so we cannot run the last released version using labpdfproc). also the default value for args.output_correction was wrong.


src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -134,11 +141,20 @@ def main():
...
+            absorption_correction = compute_cve(input_pattern, args.mud, args.wavelength)
+            corrected_data = apply_corr(input_pattern, absorption_correction)
+        else:
+            if args.mud > 6 or args.mud < 0.5:
+                sys.exit(
+                    "mu*D is out of the acceptable range (0.5 to 6) for fast calculation. "
+                    "Please rerun with a value within this range or use -b enable brute-force calculation. "
+                )
+            absorption_correction = fast_compute_cve(input_pattern, args.mud, args.wavelength)
+            corrected_data = apply_fast_corr(input_pattern, absorption_correction)
```
sbillinge, line 154: is there a need to have a different function for applying the correction?  Surely after computing cve it is the same function to apply it?  Why does this function care how it was computed?

sbillinge (commented):
looks good.  Please see comment.

src/diffpy/labpdfproc/labpdfprocapp.py
yucongalicechen: No it doesn't care the way it's computed (there are a lot of overlapping between fast and brute-force computation). I think I can remove the duplicated functions in fast_cve.

yucongalicechen: I removed the duplicated functions in fast_cve, so that the module only has one function that takes in a muD and compute the cve for that. I added a simple test on muD=1. I can add more tests on muDs other than 1, but this is probably not necessary?
The process of interpolating the computed cve and applying correction is taken care of by the original functions in functions.py (by passing the brute-force boolean).
In the last commit I moved around some misplaced docstrings in tools.py. I can do this on another PR if it's clearer.

sbillinge (commented):
I am not sure I like this refactoring tbh as it makes the `brute_force` and `fast_` calculations assymmetric.

I think a better design is to have different ways of computing the cve, all of which have the same signature and return the same data-types (and units) and when they are applied the "factory method" (but it can just be an if-statement) just figures out which to call.

You can decide whether we do both on the data grid (so pass in hte data-grid) or both on a 0.1 grid and do the interpolation outside, but I think this would be clearer and easier to maintain and extend (for example, if we come up with a new way of computing that we want to add.

Also, why does fast_cve have its own module?  Either put it in `functions.py` or make a module of `cve_calculators.py`, but probably the former unless it is getting too long and messy.

yucongalicechen:
> I am not sure I like this refactoring tbh as it makes the `brute_force` and `fast_` calculations assymmetric.
> 
> I think a better design is to have different ways of computing the cve, all of which have the same signature and return the same data-types (and units) and when they are applied the "factory method" (but it can just be an if-statement) just figures out which to call.
> 
> You can decide whether we do both on the data grid (so pass in hte data-grid) or both on a 0.1 grid and do the interpolation outside, but I think this would be clearer and easier to maintain and extend (for example, if we come up with a new way of computing that we want to add.
> 
> Also, why does fast_cve have its own module? Either put it in `functions.py` or make a module of `cve_calculators.py`, but probably the former unless it is getting too long and messy.

Okay, yes I agree it is confusing to define different tth grids. I put the fast_cve module into functions with an if-else statement on which method to use. Should be more readable now!

src/diffpy/labpdfproc/functions.py
```diff
@@ -1,12 +1,22 @@
...
 from diffpy.utils.scattering_objects.diffraction_objects import Diffraction_object
 
 RADIUS_MM = 1
 N_POINTS_ON_DIAMETER = 300
-TTH_GRID = np.arange(1, 141, 1)
+TTH_GRID = np.arange(1, 180.1, 0.1)
+
+# pre-computed datasets for fast calculation
+MUD_LIST = [0.5, 1, 2, 3, 4, 5, 6]
+CWD = os.path.dirname(os.path.abspath(__file__))
```
sbillinge, line 16: let's use pathlib.path for this work.

src/diffpy/labpdfproc/functions.py
```diff
@@ -27,7 +37,7 @@ def _get_grid_points(self):
         self.grid = {(x, y) for x in xs for y in ys if x**2 + y**2 <= self.radius**2}
         self.total_points_in_grid = len(self.grid)
 
-    # def get_coordinate_index(self, coordinate):  # I think we probably dont need this function?
+    # def get_coordinate_index(self, coordinate):
```
sbillinge, line 40: not sure what this does, but just delete it if it isn't used.

src/diffpy/labpdfproc/functions.py
```diff
@@ -172,9 +182,9 @@ def get_path_length(self, grid_point, angle):
         return total_distance, primary_distance, secondary_distance
 
 
-def compute_cve(diffraction_data, mud, wavelength):
+def compute_cve(diffraction_data, mud, wavelength, brute_force=False):
```
sbillinge, line 185: My idea was that we have different cve computers, so we don't need to pass `brute_force=false`.  We either call `fast_cve` or `brute_force_cve`, depending on whether brute_force is true or not, but the choice is handled in main.

both calculators need to have the same signatures.

sbillinge (commented):
please see my comments.  This seems to be heading a little bit in the wrong direction.  If you want to chat about the design we can, or if you have it figured out from my comments, that is fine too.

It is easy to fix.

src/diffpy/labpdfproc/functions.py
```diff
@@ -182,9 +172,62 @@ def get_path_length(self, grid_point, angle):
...
+        )
+    coeff_a, coeff_b, coeff_c, coeff_d, coeff_e = [
+        interpolation_function(mud) for interpolation_function in INTERPOLATION_FUNCTIONS
+    ]
+    muls = np.array(coeff_a * MULS**4 + coeff_b * MULS**3 + coeff_c * MULS**2 + coeff_d * MULS + coeff_e)
+    cve = 1 / muls
+    return cve
+
+
+def _cve_method(diffraction_data, mud, wavelength, brute_force=False):
```
sbillinge, line 218: this api will break if we extend the methods.  Better is to pass in method and give it a name, so `method="brute-force"` and so on.

For the argparser maintain a global list `CVE_METHODS=["brute_force", "polynomial_interpolation"]` in the top of the app main and then list these in the argparser help, so if someone types --help they can see the valid values

If we add a new method we just need to add it in three places, but the code API won't change.

src/diffpy/labpdfproc/functions.py
```diff
@@ -182,9 +172,62 @@ def get_path_length(self, grid_point, angle):
...
+    else:
+        return _cve_interp_polynomial(diffraction_data, mud, wavelength)
+
+
 def compute_cve(diffraction_data, mud, wavelength, brute_force=False):
     """
-    compute the cve for given diffraction data, mud and wavelength, and a boolean to determine the way to compute
+    compute the cve for given diffraction data, mud, and wavelength, using the selected method
 
     Parameters
```
sbillinge, line 232: Don't forget to update the docstring if the signature changes

src/diffpy/labpdfproc/functions.py
```diff
@@ -195,42 +238,13 @@ def compute_cve(diffraction_data, mud, wavelength, brute_force=False):
...
-            )
-        coeff_a, coeff_b, coeff_c, coeff_d, coeff_e = [
-            interpolation_function(mud) for interpolation_function in INTERPOLATION_FUNCTIONS
-        ]
-        muls = np.array(coeff_a * MULS**4 + coeff_b * MULS**3 + coeff_c * MULS**2 + coeff_d * MULS + coeff_e)
-        cve = 1 / muls
-
+    cve = _cve_method(diffraction_data, mud, wavelength, brute_force)
     orig_grid = diffraction_data.on_tth[0]
     newcve = np.interp(orig_grid, TTH_GRID, cve)
```
sbillinge, line 249: is there a reason we are doing this?  I don't remember.  We want to impose a uniform grid on the users?

src/diffpy/labpdfproc/functions.py
```diff
@@ -182,9 +172,62 @@ def get_path_length(self, grid_point, angle):
...
+    """
+    selects the appropriate CVE calculation method
+    """
+    if brute_force:
+        return _cve_brute_force(diffraction_data, mud, wavelength)
+    else:
+        return _cve_interp_polynomial(diffraction_data, mud, wavelength)
+
+
 def compute_cve(diffraction_data, mud, wavelength, brute_force=False):
```
sbillinge, line 228: This function naming is a bit confusing.  I think the cve is computed in the function you call `_cve_method`, so that should be called `_compute_cve`.     This function seems to do something else, like take a computed cve and put it into a diffraction object.

Let's maybe have a quick conversation about this.

sbillinge (commented):
please see comments

src/diffpy/labpdfproc/functions.py
yucongalicechen: Yes, sorry, I got this now. I'll correct this in the next commit.

src/diffpy/labpdfproc/functions.py
yucongalicechen: Yes, we first calculate the cve values on the global tth we defined, and then interpolate these values to the tth given by the user.

src/diffpy/labpdfproc/functions.py
yucongalicechen: Does `_interpolate_cve` sound good?

yucongalicechen: I added the method argument to args and cleaned up the four relevant functions in functions.py (_cve_brute_force, _cve_polynomial_interpolation, _compute_cve, and interpolate_cve).

sbillinge: > I added the method argument to args and cleaned up the four relevant functions in functions.py (_cve_brute_force, _cve_polynomial_interpolation, _compute_cve, and interpolate_cve).

I guess my question is whether we need this function.  I feel like I want a function where I give it a grid (in the form of a DO) and it gives me a cve, and a function that I give a diffraction pattern and a cve on hte same grid as each other and it gives me a corrected pattern, regardless of what backend is used.  So the interpolation step is not needed for the brute-force case, and it probably should be done as part of the cve calculation in the fast cve calculator?   I am interested in the cleanest interface for the user.

yucongalicechen: I think this function is necessary. If we use the original grid for brute-force calculations, it could take forever to run (so I think the interpolation would be necessary), and the interpolation is accurate since the numbers are small.
If we're interested, I can explore whether we can write cve as a continuous function of angle so that we can remove the interpolation?

sbillinge: that's ok, I don't mind having an interpolation. I just don't like the current structure because it gives a messy api and difficult to follow code.  btw, we might be able to interpolate onto different grids in the DO itself.   I wanted to have that capability but not sure if I implemented it.

yucongalicechen:
I updated the functions so that they now pass around DOs. There are some duplicated codes but I think it might good to have some comments on the current structure first.

yucongalicechen:
@sbillinge I think I addressed all comments here in PR #89, #92, and #93, except that there's one comment about the posixPath I haven't got to.
We have `output_dir=Path(...).resolve()` so it is the absolute Path, so in the docstring I'll correct it as a Path object.
I want to make sure that we're happy with having output directory as a Path object instead of a string object?

yucongalicechen:
Yes I think Path works better than string. I edited the docstring in #96.

yucongalicechen:
closed by PR #89, #92, #93, and #96.

PR #97, opened 2024-09-06
add back background description and an example of instruction in README.rst

yucongalicechen:
I'm adding back the background description and an example of running the app from the previous README.rst. They can be found in #84.


sbillinge:
Thanks

PR #99, opened 2024-09-09
Move tests folder

### LINKED ISSUE #98: Move tests folder at the root of the repo
see https://github.com/Billingegroup/cookiecutter/issues/114 for more details

yucongalicechen:
closes #98


bobleesj:
@yucongalicechen 

The CI fails because the installed package is unable to find example data.

Please add a line `MANIFEST.in` to import the test data as well.

From

```
recursive-exclude * __pycache__
```

to

```
recursive-include src/diffpy/labpdfproc/data *
recursive-exclude * __pycache__
```

bobleesj:
So, to mimic the CI behavior, you can run `pytest -vv` after `pip install .` instead of `pip install -e .`. 

If you use `pip install .`, you can confirm that the you are running the tests from the installed package located in `miniconda3`, etc.

yucongalicechen:
thanks for the comment! @bobleesj

bobleesj:
@yucongalicechen Just a quick heads up that we might do a quick re-cookiecutter for all diffpy projects we have all the issues addressed here: https://github.com/Billingegroup/cookiecutter/milestone/6

Since all the tests are running smoothly at the top directory level, it will be very easy!

yucongalicechen:
I'm adding back labpdfproc command here because I notice it disappears again on our main branch, but I remembered adding it back in one of the previous PRs... not sure what happened.
@sbillinge I think this PR is ready for review, thanks!

sbillinge (commented):
It looks good.  We should be able to delete __init__.py from the tests folder.  Please can you delete that and see if everything still passes?

Then I think it is ready, it looks good.

yucongalicechen:
Deleted! :)

PR #102, opened 2024-09-25
Redo cookiecutter

yucongalicechen:
Redo cookiecutter -- I think it's ready for review.


bobleesj:
thanks! @yucongalicechen

PR #108, opened 2024-10-08
Add doi

Filippo1598:
I think It is better to add doi when citing articles


bobleesj:
@Filippo1598 @yucongalicechen I believe we have no DOI at the moment. Could you please close this PR and make an issue for this? For PRs, we want to only make changes that are final since it's user-facing.

Issue #94, opened 2024-09-05
compute_cve replace the interpolation step


yucongalicechen:
- replace the interpolation step with one using diffraction objects if that ever gets implemented in diffpy.utils (https://github.com/diffpy/diffpy.utils/issues/48)
- see #92 for more detail

sbillinge:
@yucongalicechen please review and close if necessary

PR #120, opened 2024-10-31
Implement xtype other than tth

### LINKED ISSUE #45: implement other x-axes
implement x-axes other than tth grid

yucongalicechen:
closes #45 

I implemented functions to allow corrections on grid other than tth. The edits include:
- In the main module, replace all "tth" with args.xtype.
- In the tools module, filter out invalid xtype or set args.xtype to "tth" or "q" based on inputs (this makes it easy to write the output files using the dump function).
- In compute_cve function in functions.py, interpolate cve_do we get from the global tth to the input grid.

Question: I didn't see functions that allow transfer between tth and d, so when xtype=d do we simply put xtype=tth?
@sbillinge ready for some feedback.

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -16,7 +16,9 @@ from diffpy.labpdfproc.tools import (
     set_mud,
     set_output_directory,
     set_wavelength,
+    set_xtype,
 )
+from diffpy.utils.scattering_objects.diffraction_objects import XQUANTITIES
 
 # Use cases can be found here: https://github.com/diffpy/diffpy.labpdfproc/issues/48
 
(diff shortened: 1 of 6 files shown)


sbillinge:
> Question: I didn't see functions that allow transfer between tth and d, so when xtype=d do we simply put xtype=tth? @sbillinge ready for some feedback.
>
This still needs to be implemented in diffraction_objects.  Please could you make an issue if it is not already there?  d = 2*pi/Q.

sbillinge (commented):
this looks good.  It may need a test for when an invalid input is provided for xtype.  This can go in a separate test `test_set_xtype_bad`

sbillinge:
@yucongalicechen this code is very high quality.  Great job!  you have learned well!

tests/test_tools.py
```diff
@@ -189,6 +191,31 @@ def test_set_wavelength_bad(inputs, msg):
...
+
+@pytest.mark.parametrize("inputs, expected", params4)
+def test_set_xtype(inputs, expected):
+    cli_inputs = ["2.5", "data.xy"] + inputs
+    actual_args = get_args(cli_inputs)
+    actual_args = set_xtype(actual_args)
+    assert actual_args.xtype == expected[0]
+
+
+def test_set_xtype_bad():
```
yucongalicechen, line 210: Here's the test function for bad cases :)

yucongalicechen:
> > Question: I didn't see functions that allow transfer between tth and d, so when xtype=d do we simply put xtype=tth? @sbillinge ready for some feedback.
> 
> @yucongalicechen this still needs to be implemented in diffraction_objects. Please could you make an issue if it is not already there? d = 2*pi/Q.

@sbillinge 
Sounds good, I’ll add an issue in diffpy.utils.
A related question is that the dump function only writes tth and q - do we want to add q as well?
Another related question - I think we would like to have a function in diffpy.utils that interpolate DOs (as described in #94). Is this the issue here (https://github.com/diffpy/diffpy.utils/issues/48), or shall I open another issue for this?

sbillinge:
> > > Question: I didn't see functions that allow transfer between tth and d, so when xtype=d do we simply put xtype=tth? @sbillinge ready for some feedback.
> > 
> > 
> > @yucongalicechen this still needs to be implemented in diffraction_objects. Please could you make an issue if it is not already there? d = 2*pi/Q.
> 
> @sbillinge Sounds good, I’ll add an issue in diffpy.utils. A related question is that the dump function only writes tth and q - do we want to add q as well? Another related question - I think we would like to have a function in diffpy.utils that interpolate DOs (as described in #94). Is this the issue here ([diffpy/diffpy.utils#48](https://github.com/diffpy/diffpy.utils/issues/48)), or shall I open another issue for this?

yes, an issue for a user-supplied master-grid is already there i think.  Yes, dump dumping d might be nice

PR #123, opened 2024-10-31
remove wavelength from loading metadata

### LINKED ISSUE #121: Remove duplicated wavelength from output files
### Problem

Output file currently writes wavelength twice: the dump function in diffpy.utils writes the wavelength explicitly, and wavelength is written again as part of the metadata.

### Proposed solution

Remove wavelength from metadata temporarily before writing output file.

yucongalicechen:
closes #121 

Output before removal:
image2

Output after removal:
image1

@sbillinge ready for review

### DIFF
--- a/news/output-wavelength.rst
+++ b/news/output-wavelength.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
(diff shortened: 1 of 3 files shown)


sbillinge (approved):
This is good.  But I think it is non-intuitive that it would be this way, so I would add a comment on the `METADATA_KEYS_TO_EXCLUDE` that mentions  that we exclude wavelength because of (why).  this will help avoid confusion in the future likely

yucongalicechen:
@sbillinge ready for review

PR #126, opened 2024-11-06
add xraydb function to compute muD

### LINKED ISSUE #9: muD lookup table
expected behavior:
user specifies sample composition.  Given known wavelength the program computes a best effort muD.     at this point make muD an optional argument.  if muD and composition are given muD = args.mud, but composition is written to metadata.

If possible reuse other code for parsing/handling composition.  We are doing this in getx3 already.  This cannot be a dependency because it is proprietary, so maybe pull that code out into diffpy.utils?

yucongalicechen:
closes #9

I have added optional arguments (mu, d, sample, energy, density) that can be used with the xraydb function to compute muD. I didn't put any error messages for invalid inputs, because I think both the input function and xraydb give clear enough error messages (e.g. for invalid float number or incorrect sample names), but I can add some if needed.

@sbillinge ready for review

### DIFF
--- a/src/diffpy/labpdfproc/tools.py
+++ b/src/diffpy/labpdfproc/tools.py
@@ -1,6 +1,8 @@
 import copy
 from pathlib import Path
 
+from xraydb import material_mu
+
 from diffpy.labpdfproc.mud_calculator import compute_mud
 from diffpy.utils.scattering_objects.diffraction_objects import QQUANTITIES, XQUANTITIES
 from diffpy.utils.tools import get_package_info, get_user_info
@@ -158,9 +160,9 @@ def set_xtype(args):
     return args
 
 
-def set_mud(args):
+def set_mud_from_zscan_file(args):
     """
-    Set the mud based on the given input arguments
+    Set the mud based on the given z-scan file
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


src/diffpy/labpdfproc/tools.py
```diff
@@ -180,6 +182,49 @@ def set_mud(args):
...
+    the updated args argparse.Namespace with mu*D
+    """
+
+    if args.mud:
+        return args
+    if args.mu:
+        args.diameter = args.diameter or float(input("Please enter the capillary diameter (mm): ").strip())
+        args.mud = args.mu * args.diameter
+        return args
+    args.sample = args.sample or input("Please enter the chemical formula or name of material: ").strip()
```
yucongalicechen, line 219: I added some input prompts because it might be easier to enter from here than from cli inputs.

src/diffpy/labpdfproc/tools.py
```diff
@@ -281,7 +326,8 @@ def preprocessing_args(args):
     args.output_directory = set_output_directory(args)
     args = set_wavelength(args)
     args = set_xtype(args)
-    args = set_mud(args)
+    args = set_mud_from_zscan_file(args)
```
yucongalicechen, line 329: Here I am thinking that during the pre-processing we first check for z-scan files and read muD from there. If there's no file /muD provided, we then proceed to ask for other information.

yucongalicechen:
@till-schertenleib I could use some advice on this implementation, if you find other user-friendly muD functions or have any suggestions for improving the instructions!

sbillinge:
When we add user functionality like this it is important to get it right, which includes users.  Let's discuss first what use-case we want to code up before coding anything or we risk making the code worse.  @yucongalicechen do you remember how to do a Use case?  It is a list of actions that begins with an actor, something like

1. visitor wants to greet colleague politely
2. visitor holds out flat palm and approaches colleague
3. visitor stops 3 feet away and awaits response
4. colleague reaches and and grasps visitor hand 
5. visitor makes firm but not too firm grip
6. visitor moves hand up and down four times and then lets go, making eye contact and smiling
7. visitor takes a step back and starts converation.

Use your judgement, but talk to people who might use the code.  Don't start from the code.  e.g., don't say things like "user enters density" because you are then writing the UC to the code, not the code to the UC.  For example, a good UC would be

1. user places sample on diffractometer
2. diffractometer determines weight and dimensions of the sample
3. diffractometer shares this information with labpdfproc
4. labpdfproc computes muD

would be a great UC!  It may be too difficult for us to engineer right now, so we save it for later and write another UC.  In the end we impement the one that is the best balance between what is desirable and what is possible.

yucongalicechen:
> When we add user functionality like this it is important to get it right, which includes users. Let's discuss first what use-case we want to code up before coding anything or we risk making the code worse. @yucongalicechen do you remember how to do a Use case? It is a list of actions that begins with an actor, something like
> 
> 1. visitor wants to greet colleague politely
> 2. visitor holds out flat palm and approaches colleague
> 3. visitor stops 3 feet away and awaits response
> 4. colleague reaches and and grasps visitor hand
> 5. visitor makes firm but not too firm grip
> 6. visitor moves hand up and down four times and then lets go, making eye contact and smiling
> 7. visitor takes a step back and starts converation.
> 
> Use your judgement, but talk to people who might use the code. Don't start from the code. e.g., don't say things like "user enters density" because you are then writing the UC to the code, not the code to the UC. For example, a good UC would be
> 
> 1. user places sample on diffractometer
> 2. diffractometer determines weight and dimensions of the sample
> 3. diffractometer shares this information with labpdfproc
> 4. labpdfproc computes muD
> 
> would be a great UC! It may be too difficult for us to engineer right now, so we save it for later and write another UC. In the end we impement the one that is the best balance between what is desirable and what is possible.

Yes, I remember the UC. Thanks for the suggestions, I will talk to people and work on the UCs first!

yucongalicechen:
Closing this now - replaced by https://github.com/diffpy/diffpy.utils/pull/278. The UCs have been added.

PR #134, opened 2024-11-21
Use `pip install` in GitHub CI to support `Gooey` for Python 3.13 conda env

bobleesj:
The current PR https://github.com/diffpy/diffpy.labpdfproc/pull/130 fails since it attempts to `conda install` the Gooey package. However, Gooey's latest conda package support up to py3.10 while our CI setups a py3.13 env.

### Constraint
- We do not want to downgrade `labpdfproc` Python support

### Solution
- Download Gooey via pip package (pure Python) which supports Py3.13

### Impact
- This means that for `labpdfproc`, we might have to change the README instructions on how to install labpdfproc.

After this PR is merged  I will create an GH issue here to ensure we use the reusable CI workflows once Gooey supports 3.13

### DIFF
--- a/.github/workflows/matrix-and-codecov-on-merge-to-main.yml
+++ b/.github/workflows/matrix-and-codecov-on-merge-to-main.yml
@@ -11,11 +11,51 @@ on:
   workflow_dispatch:
 
 jobs:
-  matrix-coverage:
-    uses: Billingegroup/release-scripts/.github/workflows/_matrix-and-codecov-on-merge-to-main.yml@v0
-    with:
-      project: diffpy.labpdfproc
-      c_extension: false
-      headless: false
-    secrets:
-      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+  coverage:
+    defaults:
+      run:
+        shell: bash -l {0}
+
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


.github/workflows/tests-on-pr.yml
```diff
@@ -8,11 +8,46 @@ on:
...
+
+      - name: Conda config
+        run: >-
+          conda config --set always_yes yes
+          --set changeps1 no
+
+      - name: Install diffpy.labpdfproc and requirements
+        run: |
+          conda install --file requirements/test.txt
+          pip install -r requirements/pip.txt
```
bobleesj, line 38: `pip install` under `pip.txt`

.github/workflows/matrix-and-codecov-on-merge-to-main.yml
```diff
@@ -11,11 +11,51 @@ on:
...
+
+      - name: Conda config
+        run: >-
+          conda config --set always_yes yes
+          --set changeps1 no
+
+      - name: Install diffpy.labpdfproc and requirements
+        run: |
+          conda install --file requirements/test.txt
+          pip install -r requirements/pip.txt
```
bobleesj, line 48: `pip install` under `pip.txt`

PR #137, opened 2024-11-27
Support 3.12 CI - conda install `wxPython`, pip install `gooey` - to prevent building `wxPython`

bobleesj:


bobleesj:
(Not ready for review)

.github/workflows/tests-on-pr.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       run:
         shell: bash -l {0}
 
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
```
bobleesj, line 16: It turns out `wxpython` needs to be built for linux since no native whl is provided while we have windows/macos whl files provided shown below:

Screenshot 2024-11-27 at 4 03 15 PM

Since @yucongalicechen and I test using macOS, what if we setup a MacOS CI instead?

Ref:
https://pypi.org/project/wxPython/#files

requirements/conda.txt
```diff
@@ -2,4 +2,3 @@ numpy
 diffpy.utils
 pandas
 scipy
-gooey
```
bobleesj, line 5: Removed gooey from `conda.txt` since we won't be able to install it and it will only fail the CI during conda install.

bobleesj:
@sbillinge ready for review. pls see in-line comments. Thank you.

bobleesj:
@yucongalicechen tagging Yucong as well to share what's going on here.

sbillinge (commented):
Does adding wxpython to conda.txt allow you to set it back to Linux-latest?

bobleesj:
> Does adding wxpython to conda.txt allow you to set it back to Linux-latest?

I checked wxpython's conda-forge files, no support for py313 as of now:

Screenshot 2024-11-28 at 8 44 26 AM

So it will most likely break the CI while installing it. 

Ref: https://anaconda.org/conda-forge/wxpython/files

requirements/conda.txt
sbillinge: I think we need to add wxpython to conda.txt

.github/workflows/tests-on-pr.yml
sbillinge: Move back to Linux after we add wxpython to conda.txt

.github/workflows/tests-on-pr.yml
```diff
@@ -32,15 +32,11 @@ jobs:
           conda config --set always_yes yes
           --set changeps1 no
 
-      - name: Install libgtk for Linux
```
sbillinge, line 35: Restore

sbillinge (commented):
It might be cleaner to close this and redo following the regolith protocol

.github/workflows/tests-on-pr.yml
bobleesj: `libgtk` is no longer needed. here, wxpython is install via conda-forge as suggested.

requirements/conda.txt
bobleesj: yes, done.

bobleesj:
@sbillinge ready for review - thanks for the suggestion. wxpython is conda-installed as suggested while gooey is pip installed.

sbillinge:
nice.  Much cleaner.

PR #146, opened 2024-12-26
finalize `mud_calculator.py`

### LINKED ISSUE #122: Change slope bounds and number of loops in mud_calculator.py
### Problem

In mud_calculator.py, the estimated slope reaches maximum when intensity is large

### Proposed solution

- Increase the slope bounds
- Increasing the number of loops from 10 to 20 to achieve better results within a reasonable time frame
- Give a better docstring (change `slit_width` to `half_slit_width` to avoid confusion)

yucongalicechen:
closes #122 
@sbillinge ready for review. This PR is just some small edit in functions and docstring.

### DIFF
--- a/src/diffpy/labpdfproc/mud_calculator.py
+++ b/src/diffpy/labpdfproc/mud_calculator.py
@@ -91,12 +91,12 @@ def compute_mud(filepath):
 
     Parameters
     ----------
-    filepath: str
+    filepath : str
         The path to the z-scan file.
 
     Returns
     -------
-    mu*D: float
+    mu*D : float
         The best-fit mu*D value.
     """
     x_data, I_data = loadData(filepath, unpack=True)
--- a/src/diffpy/labpdfproc/mud_calculator.py
+++ b/src/diffpy/labpdfproc/mud_calculator.py
@@ -5,78 +5,78 @@ from scipy.signal import convolve
 from diffpy.utils.parsers.loaddata import loadData
 
 
-def _top_hat(x, half_slit_width):
+def _top_hat(z, half_slit_width):
     """
-    create a top-hat function, return 1.0 for values within the specified slit width and 0 otherwise
+    Create a top-hat function, return 1.0 for values within the specified slit width and 0 otherwise
     """
-    return np.where((x >= -half_slit_width) & (x <= half_slit_width), 1.0, 0)
+    return np.where((z >= -half_slit_width) & (z <= half_slit_width), 1.0, 0.0)
 
 
-def _model_function(x, diameter, x0, I0, mud, slope):
+def _model_function(z, diameter, z0, I0, mud, slope):
     """
-    compute the model function with the following steps:
-    1. Recenter x to h by subtracting x0 (so that the circle is centered at 0 and it is easier to compute length l)
+    Compute the model function with the following steps:
+    1. Recenter z to x by subtracting z0 (so that the circle is centered at 0 and it is easier to compute length l)
     2. Compute length l that is the effective length for computing intensity I = I0 * e^{-mu * l}:
-    - For h within the diameter range, l is the chord length of the circle at position h
-    - For h outside this range, l = 0
-    3. Apply a linear adjustment to I0 by taking I0 as I0 - slope * x
+    - For x within the diameter range, l is the chord length of the circle at position x
+    - For x outside this range, l = 0
+    3. Apply a linear adjustment to I0 by taking I0 as I0 - slope * z
     """
     min_radius = -diameter / 2
     max_radius = diameter / 2
-    h = x - x0
+    x = z - z0
     length = np.piecewise(
-        h,
-        [h < min_radius, (min_radius <= h) & (h <= max_radius), h > max_radius],
-        [0, lambda h: 2 * np.sqrt((diameter / 2) ** 2 - h**2), 0],
+        x,
+        [x < min_radius, (min_radius <= x) & (x <= max_radius), x > max_radius],
+        [0, lambda x: 2 * np.sqrt((diameter / 2) ** 2 - x**2), 0],
     )
-    return (I0 - slope * x) * np.exp(-mud / diameter * length)
+    return (I0 - slope * z) * np.exp(-mud / diameter * length)
 
 
-def _extend_x_and_convolve(x, diameter, half_slit_width, x0, I0, mud, slope):
+def _extend_z_and_convolve(z, diameter, half_slit_width, z0, I0, mud, slope):
     """
-    extend x values and I values for padding (so that we don't have tails in convolution), then perform convolution
+    extend z values and I values for padding (so that we don't have tails in convolution), then perform convolution
     (note that the convolved I values are the same as modeled I values if slit width is close to 0)
     """
-    n_points = len(x)
-    x_left_pad = np.linspace(x.min() - n_points * (x[1] - x[0]), x.min(), n_points)
-    x_right_pad = np.linspace(x.max(), x.max() + n_points * (x[1] - x[0]), n_points)
-    x_extended = np.concatenate([x_left_pad, x, x_right_pad])
-    I_extended = _model_function(x_extended, diameter, x0, I0, mud, slope)
-    kernel = _top_hat(x_extended - x_extended.mean(), half_slit_width)
+    n_points = len(z)
+    z_left_pad = np.linspace(z.min() - n_points * (z[1] - z[0]), z.min(), n_points)
+    z_right_pad = np.linspace(z.max(), z.max() + n_points * (z[1] - z[0]), n_points)
+    z_extended = np.concatenate([z_left_pad, z, z_right_pad])
+    I_extended = _model_function(z_extended, diameter, z0, I0, mud, slope)
+    kernel = _top_hat(z_extended - z_extended.mean(), half_slit_width)
     I_convolved = I_extended  # this takes care of the case where slit width is close to 0
     if kernel.sum() != 0:
         kernel /= kernel.sum()
         I_convolved = convolve(I_extended, kernel, mode="same")
-    padding_length = len(x_left_pad)
+    padding_length = len(z_left_pad)
     return I_convolved[padding_length:-padding_length]
 
 
-def _objective_function(params, x, observed_data):
+def _objective_function(params, z, observed_data):
     """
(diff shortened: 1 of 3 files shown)


src/diffpy/labpdfproc/mud_calculator.py
```diff
@@ -5,11 +5,11 @@
 from diffpy.utils.parsers.loaddata import loadData
 
 
-def _top_hat(x, slit_width):
+def _top_hat(x, half_slit_width):
```
yucongalicechen, line 8: change `slit_width` to `half_slit_width` to avoid confusion

src/diffpy/labpdfproc/mud_calculator.py
```diff
@@ -68,33 +68,37 @@ def _compute_single_mud(x_data, I_data):
     """
     bounds = [
         (1e-5, x_data.max() - x_data.min()),  # diameter: [small positive value, upper bound]
-        (0, (x_data.max() - x_data.min()) / 2),  # slit width: [0, upper bound]
+        (0, (x_data.max() - x_data.min()) / 2),  # half slit width: [0, upper bound]
         (x_data.min(), x_data.max()),  # x0: [min x, max x]
         (1e-5, I_data.max()),  # I0: [small positive value, max observed intensity]
         (1e-5, 20),  # muD: [small positive value, upper bound]
-        (-10000, 10000),  # slope: [lower bound, upper bound]
+        (-100000, 100000),  # slope: [lower bound, upper bound]
```
yucongalicechen, line 75: change slope bound. I think we want to support this for larger muD and larger slit widths.

src/diffpy/labpdfproc/mud_calculator.py
```diff
@@ -68,33 +68,37 @@ def _compute_single_mud(x_data, I_data):
...
 
 
 def compute_mud(filepath):
-    """
-    compute the best-fit mu*D value from a z-scan file
+    """Compute the best-fit mu*D value from a z-scan file, removing the sample holder effect.
+
+    This function loads z-scan data and fits it to a model
+    that convolves a top-hat function with I = I0 * e^{-mu * l}.
+    The fitting procedure is run multiple times, and we return the best-fit parameters based on the lowest rmse.
```
yucongalicechen, line 90: add description to docstring

src/diffpy/labpdfproc/mud_calculator.py
```diff
@@ -68,33 +68,37 @@ def _compute_single_mud(x_data, I_data):
...
 
     Returns
     -------
-    a float contains the best-fit mu*D value
+    mu*D: float
+        The best-fit mu*D value.
     """
     x_data, I_data = loadData(filepath, unpack=True)
-    best_mud, _ = min((_compute_single_mud(x_data, I_data) for _ in range(10)), key=lambda pair: pair[1])
+    best_mud, _ = min((_compute_single_mud(x_data, I_data) for _ in range(20)), key=lambda pair: pair[1])
```
yucongalicechen, line 103: change number of loops from 10 to 20

src/diffpy/labpdfproc/mud_calculator.py
```diff
@@ -32,7 +32,7 @@ def _model_function(x, diameter, x0, I0, mud, slope):
     return (I0 - slope * x) * np.exp(-mud / diameter * length)
 
 
-def _extend_x_and_convolve(x, diameter, slit_width, x0, I0, mud, slope):
+def _extend_x_and_convolve(x, diameter, half_slit_width, x0, I0, mud, slope):
```
sbillinge, line 35: I think these `x`'s should be `z`'s right?

src/diffpy/labpdfproc/mud_calculator.py
sbillinge: add a reference to your paper.

Also, I wonder if we should move this function out to `diffpy.utils.tools`?

sbillinge (commented):
please see inline comments.  Since we have the theoretical mu calculator in diffpy.utils, I wonder whether we should move the z-scan mu calculator there too?

yucongalicechen:
Yeah we can move it - I think this'll be more convenient. Shall I move it under tools or should we add a new module for these two functions?

sbillinge:
> Yeah we can move it - I think this'll be more convenient. Shall I move it under tools or should we add a new module for these two functions?

Let's put them in tools for now.  They are "tools".  We can break them out later if we want.

sbillinge:
Let's finish this PR and merge it, then make another PR that moves it out, which is just a cut and paste.

yucongalicechen:
@sbillinge ready for review

src/diffpy/labpdfproc/mud_calculator.py
```diff
@@ -5,96 +5,105 @@
 from diffpy.utils.parsers.loaddata import loadData
 
 
-def _top_hat(x, slit_width):
+def _top_hat(z, half_slit_width):
```
yucongalicechen, line 8: change all x to z

src/diffpy/labpdfproc/mud_calculator.py
```diff
@@ -5,96 +5,105 @@
...
-    - For h outside this range, l = 0
-    3. Apply a linear adjustment to I0 by taking I0 as I0 - slope * x
+    - For x within the diameter range, l is the chord length of the circle at position x
+    - For x outside this range, l = 0
+    3. Apply a linear adjustment to I0 by taking I0 as I0 - slope * z
     """
     min_radius = -diameter / 2
     max_radius = diameter / 2
-    h = x - x0
+    x = z - z0
```
yucongalicechen, line 26: change h to x to avoid confusion, since we used h as full slit width in the paper

src/diffpy/labpdfproc/mud_calculator.py
```diff
@@ -5,96 +5,105 @@
...
+        x,
+        [x < min_radius, (min_radius <= x) & (x <= max_radius), x > max_radius],
+        [0, lambda x: 2 * np.sqrt((diameter / 2) ** 2 - x**2), 0],
     )
-    return (I0 - slope * x) * np.exp(-mud / diameter * length)
+    return (I0 - slope * z) * np.exp(-mud / diameter * length)
 
 
-def _extend_x_and_convolve(x, diameter, slit_width, x0, I0, mud, slope):
+def _extend_z_and_convolve(z, diameter, half_slit_width, z0, I0, mud, slope):
```
yucongalicechen, line 35: change x to z, x0 to z0, and slit width to half slit width

src/diffpy/labpdfproc/mud_calculator.py
```diff
@@ -5,96 +5,105 @@
...
+    """Compute the best-fit mu*D value from a z-scan file, removing the sample holder effect.
+
+    This function loads z-scan data and fits it to a model
+    that convolves a top-hat function with I = I0 * e^{-mu * l}.
+    The fitting procedure is run multiple times, and we return the best-fit parameters based on the lowest rmse.
+
+    The full mathematical details are described in the paper:
+    An ad hoc Absorption Correction for Reliable Pair-Distribution Functions from Low Energy x-ray Sources,
+    Yucong Chen, Till Schertenleib, Andrew Yang, Pascal Schouwink, Wendy L. Queen and Simon J. L. Billinge,
+    in preparation.
```
yucongalicechen, line 95: add reference

src/diffpy/labpdfproc/mud_calculator.py
```diff
@@ -5,78 +5,78 @@
...
+    return np.where((z >= -half_slit_width) & (z <= half_slit_width), 1.0, 0.0)
 
 
-def _model_function(x, diameter, x0, I0, mud, slope):
+def _model_function(z, diameter, z0, I0, mud, slope):
     """
-    compute the model function with the following steps:
-    1. Recenter x to h by subtracting x0 (so that the circle is centered at 0 and it is easier to compute length l)
+    Compute the model function with the following steps:
+    1. Recenter z to x by subtracting z0 (so that the circle is centered at 0 and it is easier to compute length l)
```
sbillinge, line 18: I don't think we are using `x` correctly here either, right?  In the paper, `z` is the vertical direction centered on the center of the beam.  `x` is the direction of the x-ray beam.  We didn't define an x0 (and don't need to) but it could be something like the center of the sample.  So I am not sure that it makes sense to say "recentering z to x..."

src/diffpy/labpdfproc/mud_calculator.py
sbillinge: please see my comment above.  We shouldn't use `x` here.   For ease of reading the code, we could just leave it everywhere as `(z-z0)`.  This is how it is in the paper I think.  If we want to define a new variable it would be better to call it `dz`, `delta_z` or `z_s` or something like that?

tests/test_mud_calculator.py
```diff
@@ -3,20 +3,20 @@
...
     file = directory / "testfile"
     with open(file, "w") as f:
-        for x, I in zip(x_data, convolved_I_data):
+        for x, I in zip(z_data, convolved_I_data):
             f.write(f"{x}\t{I}\n")
 
     expected_mud = 3
     actual_mud = compute_mud(file)
-    assert actual_mud == pytest.approx(expected_mud, rel=1e-4, abs=0.1)
+    assert actual_mud == pytest.approx(expected_mud, rel=0.01, abs=0.1)
```
sbillinge, line 22: is there a reason you are making this so loose?  In general, it doesn't give me huge confidence if we need it to be so loose t pass tests....

sbillinge (commented):
much better.  A couple of more comments.

src/diffpy/labpdfproc/mud_calculator.py
yucongalicechen: Yes I will change x to dz and change this docstring here

bobleesj:
> closes #122 @sbillinge ready for review. This PR is just some small edit in functions and docstring.

General comments for PRs, while there is an issue connected, could you pls also make titles more descriptive so that I (co-maintainers) can later identify the PR for debugging/understanding by scrolling? I like using gh CLI and great titles are great.

sbillinge:
I will merge this, but @yucongalicechen maybe we could edit the title of the closed PR.  Since we are working on "best practices" let's do that.  A good thing would be for you two to figure out what the best title would be for future maintainers... So we can learn best practices for docstring titles...

PR #147, opened 2024-12-28
rename x to dz in `mud_calculator.py` and decrease error tolerance for tests

yucongalicechen:
addressed comments in #146
@sbillinge ready for review
@bobleesj let me know if this title is more informative!

### DIFF
--- a/src/diffpy/labpdfproc/mud_calculator.py
+++ b/src/diffpy/labpdfproc/mud_calculator.py
@@ -15,19 +15,19 @@ def _top_hat(z, half_slit_width):
 def _model_function(z, diameter, z0, I0, mud, slope):
     """
     Compute the model function with the following steps:
-    1. Recenter z to x by subtracting z0 (so that the circle is centered at 0 and it is easier to compute length l)
+    1. Let dz = z-z0, so that dz is centered at 0
     2. Compute length l that is the effective length for computing intensity I = I0 * e^{-mu * l}:
-    - For x within the diameter range, l is the chord length of the circle at position x
-    - For x outside this range, l = 0
+    - For dz within the capillary diameter, l is the chord length of the circle at position dz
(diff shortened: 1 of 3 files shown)


src/diffpy/labpdfproc/mud_calculator.py
```diff
@@ -15,19 +15,19 @@ def _top_hat(z, half_slit_width):
 def _model_function(z, diameter, z0, I0, mud, slope):
     """
     Compute the model function with the following steps:
-    1. Recenter z to x by subtracting z0 (so that the circle is centered at 0 and it is easier to compute length l)
+    1. Let dz = z-z0, so that dz is centered at 0
```
yucongalicechen, line 18: change x to dz

tests/test_mud_calculator.py
```diff
@@ -19,4 +19,4 @@ def test_compute_mud(tmp_path):
 
     expected_mud = 3
     actual_mud = compute_mud(file)
-    assert actual_mud == pytest.approx(expected_mud, rel=0.01, abs=0.1)
+    assert actual_mud == pytest.approx(expected_mud, rel=1e-4, abs=1e-3)
```
yucongalicechen, line 22: decrease the error tolerance. I think previously I changed this to 0.01 and 0.1 because I thought about the edge cases like HfO2 wire data. But for the other better data (pretty much all other real data we ran) the error is much lower. So we can have a smaller tolerance here.

bobleesj:
> addressed comments in #146
> 
> @sbillinge ready for review
> 
> @bobleesj let me know if this title is more informative!

Yes, great title! Thanks

tests/test_mud_calculator.py
sbillinge: This is a test, not code, so we set the tolerance we want for the test, not some worst case scenario I think, right?

Issue #150, opened 2025-01-03
feat: add `load_user_info` with options for orcid` and `skip_creation_config`


yucongalicechen:
- Implement `check_and_build_global_config()` to prompt for user inputs.
- User can specify `skip_creation_config` to bypass input prompts.
- Add an argument for user to provide orcid.

sbillinge:
I don't think we want to expose "skip_creation_config`, or whatever it is called, to users.  We put that there for developers who want to use this functionality in their programs but not bother users with the input prompts and handle things differently in their programs.   But I agree, we want to grab their orcid if we can.

PR #151, opened 2025-01-03
fix: update functions related to `diffpy.utils` update

### LINKED ISSUE #129: Check updates from diffpy.utils
Check for `xtype=d` and changes related to diffraction objects
Remove mud_calculator.py

yucongalicechen:
closes #129


src/diffpy/labpdfproc/functions.py
```diff
@@ -5,11 +5,11 @@
 import pandas as pd
 from scipy.interpolate import interp1d
 
-from diffpy.utils.scattering_objects.diffraction_objects import XQUANTITIES, Diffraction_object
+from diffpy.utils.diffraction_objects import XQUANTITIES, DiffractionObject
 
 RADIUS_MM = 1
 N_POINTS_ON_DIAMETER = 300
-TTH_GRID = np.arange(1, 180.1, 0.1)
+TTH_GRID = np.round(np.arange(1, 180.1, 0.1), decimals=1)
```
yucongalicechen, line 12: `np.arrange` causes the value at 180 degree to go above 180 slightly because of floating-point precision errors. I think we can round every value to 1 decimal place here.

src/diffpy/labpdfproc/tools.py
```diff
@@ -154,7 +153,9 @@ def set_xtype(args):
     """
     if args.xtype.lower() not in XQUANTITIES:
         raise ValueError(f"Unknown xtype: {args.xtype}. Allowed xtypes are {*XQUANTITIES, }.")
-    args.xtype = "q" if args.xtype.lower() in QQUANTITIES else "tth"
+    args.xtype = (
+        "q" if args.xtype.lower() in QQUANTITIES else "tth" if args.xtype.lower() in ANGLEQUANTITIES else "d"
```
yucongalicechen, line 157: include `d` as one of the xtype

src/diffpy/labpdfproc/tools.py
```diff
@@ -236,10 +237,9 @@ def load_user_info(args):
     the updated argparse Namespace with username and email inserted
 
     """
-    config = {"username": args.username, "email": args.email}
-    config = get_user_info(config)
-    args.username = config["username"]
-    args.email = config["email"]
+    config = get_user_info(owner_name=args.username, owner_email=args.email)
+    args.username = config["owner_name"]
+    args.email = config["owner_email"]
```
yucongalicechen, line 242: match what we have now with `get_user_info`: owner_name, owner_email, owner_orcid. I opened an issue to include options for orcid and skip_creation_config in #150.

yucongalicechen:
@sbillinge ready for review. This PR includes all updates from diffpy.utils. No significant functionality changes. But it's causing a conflict because I removed the `mud_calculator`.

src/diffpy/labpdfproc/functions.py
yucongalicechen: I ran the interpolation and coefficient lists using `np.arrange` though... but after rerunning a few angles I believe this rounding error is so small that it can be ignored

sbillinge: mmm., this seems like a slightly dramatic fix to a small edge-case.  Is there not a less intrusive fix for this?   Let's think about this a bit more.

src/diffpy/labpdfproc/functions.py
```diff
@@ -191,14 +191,14 @@ def _cve_brute_force(diffraction_data, mud):
...
     cve = 1 / muls
 
-    cve_do = Diffraction_object(wavelength=diffraction_data.wavelength)
-    cve_do.insert_scattering_quantity(
-        TTH_GRID,
-        cve,
-        "tth",
-        metadata=diffraction_data.metadata,
-        name=f"absorption correction, cve, for {diffraction_data.name}",
+    cve_do = DiffractionObject(
```
sbillinge, line 194: this feels better to me.  I think this change we made in the DO was probably a good idea.....

src/diffpy/labpdfproc/tools.py
sbillinge: nice!

src/diffpy/labpdfproc/tools.py
sbillinge: Make sure to run the config setter before the `get_user_info`

please make sure we have a test for when `args.username` etc. is None (also for the new orcid case`).  we probably do, but just double-check.

tests/test_functions.py
```diff
@@ -4,7 +4,7 @@
 import pytest
```
sbillinge, line 4: please can you annotate these tests following our new standards....what the intended behavior and what the cases cover.

sbillinge (commented):
exciting!  please see inline.

tests/test_functions.py
yucongalicechen: Yep, I opened a new issue #152 to fix docstring and tests

src/diffpy/labpdfproc/tools.py
yucongalicechen: Yeah! I will add this when I'm back (1/8).

sbillinge:
you will have to merge in main to resolve the conflict (or add the mud calculator back....)

src/diffpy/labpdfproc/functions.py
sbillinge: that's what I had in mind.

src/diffpy/labpdfproc/functions.py
```diff
@@ -5,11 +5,13 @@
...
 from scipy.interpolate import interp1d
 
-from diffpy.utils.scattering_objects.diffraction_objects import XQUANTITIES, Diffraction_object
+from diffpy.utils.diffraction_objects import XQUANTITIES, DiffractionObject
 
 RADIUS_MM = 1
 N_POINTS_ON_DIAMETER = 300
 TTH_GRID = np.arange(1, 180.1, 0.1)
+# Round down the last element if it's slightly above 180 due to floating point precision
+TTH_GRID[-1] = 180
```
yucongalicechen, line 14: round down 180 only

src/diffpy/labpdfproc/tools.py
yucongalicechen: I added the config setter. We have a test case where args are None.

yucongalicechen:
@sbillinge ready for another review. I fixed the `TTH_GRID` and resolved the conflicts. I will open another PR on fixing docstrings and tests for username orcid etc.

news/utils-updates.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Updated functions related to changes in `diffpy.utils` version 3.6.0.
```
sbillinge, line 7: Can you try and think more precisely about how to say this.  "Updated" is the same as "changed" so the message contains close to zero information.....

src/diffpy/labpdfproc/functions.py
sbillinge: maybe 180.00 so it is clear you want to keep it as a float?

src/diffpy/labpdfproc/tools.py
```diff
@@ -1,9 +1,8 @@
 import copy
 from pathlib import Path
 
-from diffpy.labpdfproc.mud_calculator import compute_mud
-from diffpy.utils.scattering_objects.diffraction_objects import QQUANTITIES, XQUANTITIES
-from diffpy.utils.tools import get_package_info, get_user_info
+from diffpy.utils.diffraction_objects import ANGLEQUANTITIES, QQUANTITIES, XQUANTITIES
+from diffpy.utils.tools import check_and_build_global_config, compute_mud, get_package_info, get_user_info
 
 WAVELENGTHS = {"Mo": 0.71, "Ag": 0.59, "Cu": 1.54}
```
sbillinge, line 7: I think we want a few more sig figs on these

src/diffpy/labpdfproc/tools.py
```diff
@@ -236,10 +237,10 @@ def load_user_info(args):
     the updated argparse Namespace with username and email inserted
 
     """
-    config = {"username": args.username, "email": args.email}
-    config = get_user_info(config)
-    args.username = config["username"]
-    args.email = config["email"]
+    check_and_build_global_config()
```
sbillinge, line 240: since here we are allowing people to pass in this info as an arg, probably the workflow should be

- if the arg is set, use the arg and don't even run the config updater
- if the arg is not set, run the config updater
- run get_user_info passing in the args if they are not None.

-

src/diffpy/labpdfproc/tools.py
```diff
@@ -236,10 +237,10 @@ def load_user_info(args):
...
+    config = get_user_info(owner_name=args.username, owner_email=args.email)
+    args.username = config["owner_name"]
```
sbillinge, line 242: is there a reason you are not using `config.get()` here?  That is usually better practice.

sbillinge (commented):
I didn't quite finish the review and have to run, but I am posting this now so you can work on it.

src/diffpy/labpdfproc/tools.py
yucongalicechen: I found for Mo and Cu the values should be around 0.7107 and 1.5406. I couldn't find the values for Ag though..

news/utils-updates.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Functions related to diffraction_objects and tools modules in `diffpy.utils` to reflect changes in version 3.6.0.
```
yucongalicechen, line 7: better wording..?

src/diffpy/labpdfproc/functions.py
```diff
@@ -5,11 +5,13 @@
...
 from scipy.interpolate import interp1d
 
-from diffpy.utils.scattering_objects.diffraction_objects import XQUANTITIES, Diffraction_object
+from diffpy.utils.diffraction_objects import XQUANTITIES, DiffractionObject
 
 RADIUS_MM = 1
 N_POINTS_ON_DIAMETER = 300
 TTH_GRID = np.arange(1, 180.1, 0.1)
+# Round down the last element if it's slightly above 180 due to floating point precision
+TTH_GRID[-1] = 180.00
```
yucongalicechen, line 14: changed from 180 to 180.00

src/diffpy/labpdfproc/tools.py
```diff
@@ -1,11 +1,10 @@
...
 from pathlib import Path
 
-from diffpy.labpdfproc.mud_calculator import compute_mud
-from diffpy.utils.scattering_objects.diffraction_objects import QQUANTITIES, XQUANTITIES
-from diffpy.utils.tools import get_package_info, get_user_info
+from diffpy.utils.diffraction_objects import ANGLEQUANTITIES, QQUANTITIES, XQUANTITIES
+from diffpy.utils.tools import check_and_build_global_config, compute_mud, get_package_info, get_user_info
 
-WAVELENGTHS = {"Mo": 0.71, "Ag": 0.59, "Cu": 1.54}
+WAVELENGTHS = {"Mo": 0.71073, "Ag": 0.55941, "Cu": 1.5406}
```
yucongalicechen, line 7: Need a little help on the value for Ag. I found it here https://x-server.gmca.aps.anl.gov/cgi/www_dbli.exe?x0hdb=waves but it's a bit off from our original value..

yucongalicechen:
@sbillinge ready for some more feedback. The error in the tests is related to the wavelength for Ag.

news/utils-updates.rst
sbillinge: not really.  None of the users reading this will have any idea about diffpy.utils 3.6.0.

Something like 
```
* Functions that use DiffractionObject` in `diffpy.utils` to follow the new API.
```

src/diffpy/labpdfproc/tools.py
sbillinge: "Cu", "CuKa1", "CuKa1Ka2" 
CuKa1 : 1.54056
CuKa1Ka2 : =(1.54056*2 + 1.544398)/3 = 1.54184
Cu : 1.542

sbillinge: @yucongalicechen will need docs to explain what is going on.

sbillinge (commented):
please see inline

PR #162, opened 2025-01-31
feat: new wavelength workflow

### LINKED ISSUE #145: 0.2.1 release issues overview
docs:
- [x] #114
- [x] #132
- [x] #148
- [x] #157 
- [x] #158 
- [x] #159
- [x] #171

diffpy.utils related:
- [x] #122 
- [x] #129
- [x] #150 
- [x] #152 

muD:
- [x] #143
- [x] #166 

other features:
- [x] #57 
- [x] #45 
- [x] #121 
- [x] #149 
- [x] #153
- [x] #160 
- [x] #161
- [x] #167
- [x] #109 
- [x] #174

### LINKED ISSUE #161: feat: workflow for getting wavelength/anode type
Refer to #155 for more discussion.

Try to get wavelength/anode type from args, then from config files.
C1: User specified nothing. If mu*D is given and xtype is tth, proceed. Otherwise, throw an error.
C2: User specified both, throw an error.
C3: User specified anode type. Check if it is in the database. If not throw an error.
C4: User specified wavelength. Check if it is valid. If not throw an error.

yucongalicechen:
closes #161
I will add functionality to load wavelength/anode type from config file in another PR after this is merged
@sbillinge ready for review

### DIFF
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -30,8 +30,7 @@ def define_arguments():
                 "in a file with name file_list.txt. "
                 "If one or more directory is provided, all valid "
                 "data-files in that directory will be processed. "
-                "Examples of valid "
-                "inputs are 'file.xy', 'data/file.xy', "
+                "Examples of valid inputs are 'file.xy', 'data/file.xy', "
                 "'file.xy, data/file.xy', "
                 "'.' (load everything in the current directory), "
                 "'data' (load everything in the folder ./data), "
@@ -48,18 +47,16 @@ def define_arguments():
             "name": ["-a", "--anode-type"],
             "help": (
                 f"The type of the x-ray source. "
-                f"Allowed values are {*[known_sources], }. "
+                f"Allowed values are {*known_sources, }. "
                 f"Either specify a known x-ray source or specify wavelength."
             ),
-            "default": "Mo",
+            "default": None,
         },
         {
             "name": ["-w", "--wavelength"],
             "help": (
                 "X-ray source wavelength in angstroms. "
                 "Not needed if the anode-type is specified. "
-                "This wavelength will override the anode wavelength "
-                "if both are specified."
             ),
             "type": float,
         },
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -56,7 +56,7 @@ def define_arguments():
             "name": ["-w", "--wavelength"],
             "help": (
                 "X-ray source wavelength in angstroms. "
-                "Not needed if the anode-type is specified. "
+                "Not needed if the anode-type is specified."
             ),
             "type": float,
         },
@@ -101,7 +101,7 @@ def define_arguments():
                 f"The method for computing absorption correction. "
                 f"Allowed methods: {*CVE_METHODS, }. "
                 f"Default method is polynomial interpolation "
-                f"if not specified. "
+                f"if not specified."
             ),
             "default": "polynomial_interpolation",
         },
@@ -118,7 +118,7 @@ def define_arguments():
                 "For example, facility='NSLS II', "
                 "'facility=NSLS II', beamline=28ID-2, "
                 "'beamline'='28ID-2', 'favorite color'=blue, "
-                "are all valid key=value items. "
+                "are all valid key=value items."
             ),
             "nargs": "+",
             "metavar": "KEY=VALUE",
@@ -128,7 +128,7 @@ def define_arguments():
             "help": (
                 "Username will be loaded from config files. "
                 "Specify here only if you want to "
-                "override that behavior at runtime. "
+                "override that behavior at runtime."
             ),
             "default": None,
         },
@@ -137,7 +137,7 @@ def define_arguments():
             "help": (
                 "Email will be loaded from config files. "
                 "Specify here only if you want to "
-                "override that behavior at runtime. "
+                "override that behavior at runtime."
             ),
             "default": None,
         },
@@ -146,7 +146,7 @@ def define_arguments():
             "help": (
                 "ORCID will be loaded from config files. "
                 "Specify here only if you want to "
-                "override that behavior at runtime. "
+                "override that behavior at runtime."
             ),
             "default": None,
         },
--- a/src/diffpy/labpdfproc/tools.py
+++ b/src/diffpy/labpdfproc/tools.py
@@ -175,8 +175,7 @@ def set_wavelength(args):
     ValueError
         Raised if:
         (1) neither wavelength or anode type is provided,
-            and either mu*D needs to be looked up or
-            xtype is not the two-theta grid,
+            and xtype is not the two-theta grid,
         (2) both are provided,
         (3) anode_type is not one of the known sources,
         (4) wavelength is non-positive.
@@ -188,11 +187,10 @@ def set_wavelength(args):
     """
     # first load values from config file
     if args.wavelength is None and args.anode_type is None:
-        # either mu*D needs to be looked up or
-        # xtype is not the two-theta grid
-        if args.mud is None or args.xtype not in ANGLEQUANTITIES:
+        if args.xtype not in ANGLEQUANTITIES:
             raise ValueError(
-                f"Please provide a wavelength or anode type. "
+                f"Please provide a wavelength or anode type "
+                f"because the independent variable axis is not on two-theta. "
                 f"Allowed anode types are {*known_sources, }."
             )
     elif args.wavelength is not None and args.anode_type is not None:
--- a/src/diffpy/labpdfproc/tools.py
+++ b/src/diffpy/labpdfproc/tools.py
@@ -160,9 +160,9 @@ def set_input_lists(args):
 
 
 def set_wavelength(args):
-    """Set the wavelength based on the given anode_type.
-    If a wavelength is provided,
-    it will be used, and the anode_type argument will be removed.
+    """Set the wavelength based on the given anode_type or wavelength.
+    First checks from args. If neither is provided,
+    it attempts to load from local and then global config file.
 
     Parameters
     ----------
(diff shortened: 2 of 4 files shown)


src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -48,18 +47,16 @@ def define_arguments():
             "name": ["-a", "--anode-type"],
             "help": (
                 f"The type of the x-ray source. "
-                f"Allowed values are {*[known_sources], }. "
+                f"Allowed values are {*known_sources, }. "
                 f"Either specify a known x-ray source or specify wavelength."
             ),
-            "default": "Mo",
+            "default": None,
```
yucongalicechen, line 53: change default from Mo to None

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -48,18 +47,16 @@ def define_arguments():
...
         },
         {
             "name": ["-w", "--wavelength"],
             "help": (
                 "X-ray source wavelength in angstroms. "
                 "Not needed if the anode-type is specified. "
-                "This wavelength will override the anode wavelength "
```
yucongalicechen, line 61: remove this functionality

src/diffpy/labpdfproc/tools.py
```diff
@@ -172,15 +172,31 @@ def set_wavelength(args):
...
+        (3) anode_type is not one of the known sources,
+        (4) wavelength is non-positive.
 
     Returns
     -------
     args : argparse.Namespace
         The updated arguments with the wavelength.
     """
-    if args.wavelength is None:
+    # first load values from config file
```
yucongalicechen, line 187: added a comment here. will make another PR for this functionality

tests/test_tools.py
```diff
@@ -239,45 +237,51 @@ def test_set_output_directory_bad(user_filesystem):
...
     [
-        (
+        (  # C1.1: nothing passed in, xtype is not on tth
+            # expect error asking for either wavelength or anode type
+            ["--xtype", "q"],
+            f"Please provide a wavelength or anode type. "
+            f"Allowed anode types are {*known_sources, }.",
+        ),
+        # C1.2: nothing passed in, need to look up mu*D
+        # expect error asking for either wavelength or anode type
```
yucongalicechen, line 267: I added a comment here in case we make mu*D optional later

src/diffpy/labpdfproc/tools.py
```diff
@@ -172,15 +172,31 @@ def set_wavelength(args):
     Raises
     ------
     ValueError
-        Raised when input wavelength is non-positive
-        or if input anode_type is not one of the known sources.
+        Raised if:
+        (1) neither is provided, and either mu*D needs to be looked up or
```
sbillinge, line 176: neither what?  It is implied but I would make it explicit here.

src/diffpy/labpdfproc/tools.py
sbillinge: When you do that, if possible, let's make things reusable.  Either reuse the user machinery in diffpy.utils, or refactor over there so that we can reuse it here for this more general task, rather than putting some bespoke function here.  I think one task is maintaining and interacting with a diffpy config file, and specific functions are interacting it for user, orcid and so on, and another it interact with it for wavelength, etc.  let's just come up with some "plan" that makes sense.

src/diffpy/labpdfproc/tools.py
```diff
@@ -172,15 +172,31 @@ def set_wavelength(args):
...
 
     Returns
     -------
     args : argparse.Namespace
         The updated arguments with the wavelength.
     """
-    if args.wavelength is None:
+    # first load values from config file
+    if args.wavelength is None and args.anode_type is None:
+        if not (args.mud is not None and args.xtype in ANGLEQUANTITIES):
```
sbillinge, line 189: a comment might help readability here.  Whilst the logic can, in principle, be parsed from the code, double-negatives and ANDs make it hard, so a brief comment about the desired outcome could help here.  Normally, I like code to "self comment" and not put comments, and to simplify code to accomplish this, but here probably there is no way to simplify the logic correctly so a comment is ok.

tests/test_tools.py
```diff
@@ -239,45 +237,51 @@ def test_set_output_directory_bad(user_filesystem):
...
+        # C2: a valid wavelength was entered,
         # expect to include the wavelength only and anode type is None
         (["--wavelength", "0.25"], {"wavelength": 0.25, "anode_type": None}),
-        # C4: both valid anode type and wavelength were entered,
-        # expect to remove the anode type and preserve wavelength only
-        (
-            ["--wavelength", "0.25", "--anode-type", "Ag"],
-            {"wavelength": 0.25, "anode_type": None},
-        ),
+        # C3: nothing passed in, but mu*D was provided and xtype is on tth
```
sbillinge, line 243: these are excellent by the way!  So clear!

tests/test_tools.py
```diff
@@ -239,45 +237,51 @@ def test_set_output_directory_bad(user_filesystem):
...
+        ),
+        # C1.2: nothing passed in, need to look up mu*D
+        # expect error asking for either wavelength or anode type
+        (  # C2: both wavelength and anode type were specified
+            # expect error asking not to specify both
+            ["--wavelength", "0.7", "--anode-type", "Mo"],
+            f"Please provide either a wavelength or an anode type, not both. "
+            f"Allowed anode types are {*known_sources, }.",
+        ),
+        (  # C3: invalid anode type
```
sbillinge, line 274: does this need an "expect" statement?

tests/test_tools.py
```diff
@@ -239,45 +237,51 @@ def test_set_output_directory_bad(user_filesystem):
...
             ["--anode-type", "invalid"],
             f"Anode type not recognized. "
             f"Please rerun specifying an anode_type from {*known_sources, }.",
         ),
-        (
+        (  # C4: invalid wavelength
```
sbillinge, line 279: expect

tests/test_tools.py
```diff
@@ -515,6 +519,8 @@ def test_load_metadata(mocker, user_filesystem):
     cli_inputs = [
```
sbillinge, line 519: I think this needs the "such and such, expect so and so" treatment

sbillinge (commented):
super!  Please see comments.

src/diffpy/labpdfproc/tools.py
yucongalicechen: I think the idea would be similar to the `get_user_info` function, but since the parameters we pass in are different we cannot reuse it completely. I'm thinking about a function like this:
`def load_config_file(args):
         load local config file
         load global config file
         if wavelength/anodetype is present in args:
               return
         elif look for wavelength/anode type in local config file:
               if present:
                   return
         elif look for wavelength/anode type in global config file:
               if present:
                   return
         return args`

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -30,8 +30,7 @@ def define_arguments():
                 "in a file with name file_list.txt. "
                 "If one or more directory is provided, all valid "
                 "data-files in that directory will be processed. "
-                "Examples of valid "
-                "inputs are 'file.xy', 'data/file.xy', "
+                "Examples of valid inputs are 'file.xy', 'data/file.xy', "
```
yucongalicechen, line 33: remove the whitespace here

src/diffpy/labpdfproc/tools.py
```diff
@@ -160,9 +160,9 @@ def set_input_lists(args):
 
 
 def set_wavelength(args):
-    """Set the wavelength based on the given anode_type.
-    If a wavelength is provided,
-    it will be used, and the anode_type argument will be removed.
+    """Set the wavelength based on the given anode_type or wavelength.
+    First checks from args. If neither is provided,
```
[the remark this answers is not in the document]
yucongalicechen: added

src/diffpy/labpdfproc/tools.py
```diff
@@ -172,15 +173,34 @@ def set_wavelength(args):
...
     -------
     args : argparse.Namespace
         The updated arguments with the wavelength.
     """
-    if args.wavelength is None:
+    # first load values from config file
+    if args.wavelength is None and args.anode_type is None:
+        # either mu*D needs to be looked up or
+        # xtype is not the two-theta grid
+        if args.mud is None or args.xtype not in ANGLEQUANTITIES:
```
yucongalicechen, line 193: added a comment here and simplified the logic

tests/test_tools.py
```diff
@@ -502,6 +508,11 @@ def test_load_package_info(mocker):
 
 
 def test_load_metadata(mocker, user_filesystem):
+    # Test if the function loads args
+    # (which will be loaded into the header file).
+    # Expect to include mu*D, anode type, xtype, cve method,
+    # user-specified metadata, user info, package info, z-scan file,
+    # and full paths for current input and output directories.
```
yucongalicechen, line 515: added test comment

src/diffpy/labpdfproc/tools.py
yucongalicechen: I think the difference between this and user info is that we cannot just use what's in the config file if user specified anything in args. For example, if user specified `wavelength=0.25` while having config file info with `anodetype=Mo` we don't want to overload anode type to Mo which will cause an error. So I'm not sure how much can be reused. But maybe we can make this more general here, so that this function can be reused in the future?

yucongalicechen:
@sbillinge ready for another review!

src/diffpy/labpdfproc/tools.py
sbillinge: I guess what I am saying is that there are general tasks (like `load_config_file()`) that should be in diffpy-utils and do just one thing (in that case, run the config-file loading logic).  

Then different apps like yours would use that file, so in your app you would have logic like
```
config_args = load_config_file()
parser_args = load_parser()
args = <logic for merging config and cli args>
```
kind of thing.  Your "load_config_file" is doing two things. It is loading the config files in the right order, but it is also running logic on the contents.

sbillinge (commented):
good progress.  Please see my comment. We can move things to a different PR to make it easier to review if you want to make an issue again.....

yucongalicechen:
> good progress. Please see my comment. We can move things to a different PR to make it easier to review if you want to make an issue again.....

Yes I can work on the config workflow on a separate PR. I see what you mean. I think there's a load_config function in diffpy.utils that we can reuse.

src/diffpy/labpdfproc/tools.py
```diff
@@ -172,15 +173,32 @@ def set_wavelength(args):
     Raises
     ------
     ValueError
-        Raised when input wavelength is non-positive
-        or if input anode_type is not one of the known sources.
+        Raised if:
+        (1) neither wavelength or anode type is provided,
```
yucongalicechen, line 177: I removed the muD requirement here to simplify the logic. I think we will set wavelength before setting muD, so maybe we can raise the error there when we look up the muD?

src/diffpy/labpdfproc/tools.py
```diff
@@ -172,15 +173,32 @@ def set_wavelength(args):
...
         The updated arguments with the wavelength.
     """
-    if args.wavelength is None:
+    # first load values from config file
+    if args.wavelength is None and args.anode_type is None:
+        if args.xtype not in ANGLEQUANTITIES:
+            raise ValueError(
+                f"Please provide a wavelength or anode type "
+                f"because the independent variable axis is not on two-theta. "
+                f"Allowed anode types are {*known_sources, }."
```
yucongalicechen, line 194: edited error msg

sbillinge:
ok, thanks @yucongalicechen .  I got a bit tired of looking at this PR so I merged it.  I hope that we have good issues posted for all the unfinished work...... :)

PR #163, opened 2025-01-31
docs: for functions module

### LINKED ISSUE #157: docs: functions module
subissue for #114

yucongalicechen:
closes #157 
@sbillinge ready for review


doc/source/examples/functions_example.rst
```diff
@@ -0,0 +1,74 @@
...
+    import matplotlib.pyplot as plt
+    plt.plot(input_pattern.xarray, input_pattern.yarray, label="Original Intensity")
+    plt.plot(corrected_pattern.xarray, corrected_pattern.yarray, label="Corrected Intensity")
+    plt.xlabel("tth (degrees)")
+    plt.ylabel("Intensity")
+    plt.legend()
+    plt.title("Original vs. Corrected Intensity")
+    plt.show()
+
+4. You can modify the global parameters
```
yucongalicechen, line 64: For this section, I'm wondering if I should add these parameters in the brute force computation function, e.g.,
`compute_cve(input_pattern, mud, method="polynomial_interpolation", xtype="tth", n_points_on_diameter=None, tth_grid=None)`
so that people can edit these without modifying the global parameters. This is probably uncommon so we don't need to modify anything in the main app.

sbillinge: This would make this more general and so better.  However, for things like this we may want to wait until someone asks for it and then do it?  Maybe put it as a feature request in the issues so we don't forget?  We could create a milestone which is a holding place for feature requests, and we can move them to a release milestone when someone requiests it and we decide we want to do it?

sbillinge (commented):
please see comment.  I will merge this because documentation is easier to review when it is built, but @yucongalicechen please read and act on the comment.

doc/source/examples/functions_example.rst
yucongalicechen: Sounds great, I'll do that!

yucongalicechen: It seems that I can't add a new milestone.. but I created an issue for this

Issue #158, opened 2025-01-27
docs: tools module


yucongalicechen:
subissue for #114

- [ ] instructions for writing wavelength/anodetype to config file (see https://github.com/diffpy/diffpy.labpdfproc/pull/169).

PR #169, opened 2025-02-24
feat: config workflow for wavelength / anode type

### LINKED ISSUE #160: feat: add config workflow for wavelength
I wonder if the default should be `MoKa1`.  I am not sure which is the most commonly used.  Also, it would be good if it could be loaded from a config file.  In this way, if it is lab diffractometer, the correct one could be in the config file and loaded by default, but in a different lab, they could update the config file accordingly.

_Originally posted by @sbillinge in https://github.com/diffpy/diffpy.labpdfproc/pull/155#discussion_r1931031579_

yucongalicechen:
closes #160 
@sbillinge ready for review


tests/test_tools.py
```diff
@@ -195,6 +197,121 @@ def test_set_output_directory_bad(user_filesystem):
...
+        # expect to return args if wavelength or anode type is specified,
+        # otherwise update args with values from the home config file.
+        # C1: no args, expect to update arg values from home config
+        ([""], {"wavelength": 0.3, "anode_type": None}),
+        # C2: wavelength provided, expect to return args unchanged
+        (["--wavelength", "0.25"], {"wavelength": 0.25, "anode_type": None}),
+        # C3: anode type provided, expect to return args unchanged
+        (["--anode-type", "Mo"], {"wavelength": None, "anode_type": "Mo"}),
+        # C4: both wavelength and anode type provided,
+        # expect to return args unchanged
```
sbillinge, line 213: Is this what we want?  I think it should be either `wavelength` or `anode_type`.  If both are provided raise an exception.

tests/test_tools.py
```diff
@@ -195,6 +197,121 @@ def test_set_output_directory_bad(user_filesystem):
...
+        # otherwise update args with values from the home config file.
+        # Results should be the same whether if the home config exists.
+        # C1: no args, expect to update arg values from local config
+        ([""], {"wavelength": 0.6, "anode_type": None}),
+        # C2: wavelength provided, expect to return args unchanged
+        (["--wavelength", "0.25"], {"wavelength": 0.25, "anode_type": None}),
+        # C3: anode type provided, expect to return args unchanged
+        (["--anode-type", "Mo"], {"wavelength": None, "anode_type": "Mo"}),
+        # C4: both wavelength and anode type provided,
+        # expect to return args unchanged
```
sbillinge, line 249: I agree that the tests should be the same here, as they are, but we then need to modify the last test in both places.

Also maybe mention in the preamble that the home config file has wavelength of 0.3 and the local config has wavelength of 0.6 to make the test easier to read.

A thought occurred to me that maybe having both wavelength and anode_type specified is tested elsewhere and you do want to return both values here but have it fail in a test elsewhere, but i fthat is the case, say it explicitly in the comment about the test.

tests/test_tools.py
```diff
@@ -195,6 +197,121 @@ def test_set_output_directory_bad(user_filesystem):
...
+    os.remove(confile)
+    assert actual_args.wavelength == expected["wavelength"]
+    assert actual_args.anode_type == expected["anode_type"]
+
+
+@pytest.mark.parametrize(
+    "inputs, expected",
+    [
+        # Test when no config files exist,
+        # expect to return args without modification.
```
sbillinge, line 284: here don't we expect it to run the "create config" workflow if no wavelength/anode_type is specified?

sbillinge (commented):
nice work.  Please see inline

tests/test_tools.py
yucongalicechen: Yeah I think the idea is that this function just updates the correct args. The error will be reported when we set the wavelength etc. I will add a comment here.
For the config workflow I think we discussed that we might not want to make it complicated for users?

sbillinge: I am not sure what we discussed.  We created to the "create config workflow" to be as intuitive to use as possible.  After that it should just be a choice in apps (a) use it (b) don't use it, so I am not sure I understand your statement about making things not complicated for users.

yucongalicechen: Getting back to this - I just remembered what we discussed. I think it's a bit complicated to create config workflow here because we want user to enter either wavelength or anode type but not both. So it might be better if we just let the user decide whether to put wavelength/anode type in the config file manyally. Maybe we can load it from args to the config file automatically, if that make things easier?

sbillinge: right, I see.  Yes, we could just put it in the docs.  In general, we want this behavior set up by say a lab manager (llike a beamline scientist) so that every user who runs the machine doesn't have to enter the value manually.  So a complicated command line way of capturing it is probably not warranted.  I agree.  But let's put it in the docs to merge this.

yucongalicechen: Sounds good, I added a task in issue #158 to add instructions in docs. I updated the codes to address the comments above.

src/diffpy/labpdfproc/tools.py
```diff
@@ -159,6 +160,52 @@ def set_input_lists(args):
     return args
 
 
+def load_wavelength_from_config_file(args):
```
yucongalicechen, line 163: Made the function public

src/diffpy/labpdfproc/tools.py
```diff
@@ -159,6 +160,52 @@ def set_input_lists(args):
...
+    global_config = _load_config(Path().home() / "diffpyconfig.json")
+    local_config = _load_config(Path().cwd() / "diffpyconfig.json")
+    local_has_data = local_config and (
+        "wavelength" in local_config or "anode_type" in local_config
+    )
+    global_has_data = global_config and (
+        "wavelength" in global_config or "anode_type" in global_config
+    )
+    if not local_has_data and not global_has_data:
+        print(
```
yucongalicechen, line 188: Print instructions for manually adding wavelength to config file when it's not found

tests/test_tools.py
```diff
@@ -195,6 +197,131 @@ def test_set_output_directory_bad(user_filesystem):
...
+@pytest.mark.parametrize(
+    "inputs, expected",
+    [
+        # Test with only a home config file (no local config),
+        # expect to return values directly from args
+        # if either wavelength or anode type is specified,
+        # otherwise update args with values from the home config file
+        # (wavelength=0.3, no anode type).
+        # This test only checks loading behavior,
+        # not value validation (which is handled by `set_wavelength`).
```
yucongalicechen, line 209: Added home config info about wavelength/anodetype and clarified that this function only checks loading behavior

tests/test_tools.py
```diff
@@ -279,13 +406,13 @@ def test_set_wavelength(inputs, expected):
         (  # C3: invalid anode type
             # expect error asking to specify a valid anode type
             ["--anode-type", "invalid"],
-            f"Anode type not recognized. "
+            f"Anode type 'invalid' not recognized. "
```
yucongalicechen, line 409: Updated error message slightly to be more informative

yucongalicechen:
Okay, I did some cleanup and updates to match the current muD cli behavior - ready to be review now @sbillinge

src/diffpy/labpdfproc/tools.py
```diff
@@ -174,16 +176,33 @@ def _load_wavelength_from_config_file(args):
...
+            "For more information, please refer to www.diffpy.org/"
+            "diffpy.labpdfproc/examples/toolsexample.html"
+        )
+
+    if args.wavelength or args.anode_type:
+        return args
+    config = local_config if local_has_data else global_config
+    if config:
+        args.wavelength = args.wavelength or config.get("wavelength")
+        args.anode_type = args.anode_type or config.get("anode_type")
```
sbillinge, line 205: this is all very nicely done.  I like the coding style very much.  Very readable.

sbillinge (commented):
very nicely done.  Let's user test now and make sure users like the workflow.

Issue #167, opened 2025-02-07
feat: muD fast calculation for values <0.5 and >6


yucongalicechen:
### Proposed solution

1. Include muD up to 7 for fast calculation.
2. Default to brute-force computation when muD < 0.5 or > 7; report muD value during computation.

sbillinge:
why are we going from 6 to 7?  This seems pretty arbitrary.  I think it is more a philosophical issue about how to handle this flow.  We can define `mud_threshold` for example.  Right now `mud_threshold = 6`.  We could change mud_threshold to something else, but the conversation here might be better to be about what flow we want when`mud_threshold` is exceeded in general.

PR #170, opened 2025-04-15
feat: increase fast cal to muD=7

yucongalicechen:
addressed bullet points 2 and 3 in #167
@sbillinge ready to review. I can address bullet point 1 in #167 on this PR too (it's about not raising an error for muD < 0.5), but I'm not sure what the expected output should be in that case.


src/diffpy/labpdfproc/functions.py
```diff
@@ -206,25 +206,28 @@ def _cve_brute_force(input_pattern, mud):
...
 def _cve_polynomial_interpolation(input_pattern, mud):
     """Compute cve using polynomial interpolation method,
-    raise an error if the mu*D value is out of the range (0.5 to 6).
+    raise an error if the mu*D value is out of the range (0.5 to 7).
     """
-    if mud > 6 or mud < 0.5:
+    if mud > 7 or mud < 0.5:
         raise ValueError(
-            f"mu*D is out of the acceptable range (0.5 to 6) "
+            f"Input mu*D = {mud} is out of the acceptable range "
```
sbillinge, line 213: I still think this message is a bit harsh. 

Is there a reason we don't just print a warning message and default to computing using the brute-force method?

sbillinge (commented):
please see comment.  I suggested a different way to handle this.

src/diffpy/labpdfproc/functions.py
yucongalicechen: Edited! Please check.

src/diffpy/labpdfproc/functions.py
```diff
@@ -206,28 +202,21 @@ def _cve_brute_force(input_pattern, mud):
...
-        + coeff_c * MULS**2
-        + coeff_d * MULS
-        + coeff_e
-    )
-    cve = 1 / muls
+        return _cve_brute_force(input_pattern, mud)
 
+    coeffs = np.array([f(mud) for f in INTERPOLATION_FUNCTIONS])
+    muls = np.polyval(coeffs, MULS)
+    cve = 1 / muls
```
yucongalicechen, line 219: Simplified the codes here.

sbillinge: nice!

PR #173, opened 2025-04-23
feat: options for estimating muD theoretically

### LINKED ISSUE #143: Implement multiple options for computing muD
We would like an app that reads things from the command line (or gooey), calls the functions from here, and goes on its merry way. Its tricky in gooey but we may be able to do one layer of conditional logic, to support a gooey uc like this:

- [ ] UC3a
1. user selects "compute muD from known density" radio button
2. labpdfproc.mudCalc presents a form pre-filled with defaults containing fields for user to enter mass density in g/cm^3, composition (e.g. ZrO2), wavelength(angstroms), D (mm). Defaults are <Till's zirconia number>, , , 1.
3. user enters the other info (browses for files etc.)
4. user hits start
5. code runs.

- [x] UC3b - user has muD
1. user selects "enter muD manually"
2. mudCalc presents a box to type in a float "muD value"
3. as UC3a 3-end

- [ ] UC3c - user doesn't know density
1. user selects "estimate muD from composition"
2. labpdfproc.mudCalc presents a form pre-filled with defaults containing fields for user to enter powder fraction (number between 0 and 1), composition (e.g. ZrO2), wavelength(angstroms), D (mm). Defaults are 0.5, , , 1. program assumes some reasonable atomic number density by default, since this is not known or given in principle.
3. as UC3a 3-end

- [x] UC4
1. as UC3a.1-2
2. user clicks run calc
3. lpp.mudCalc returns the calculated value as output without running the cve calc.

Implementationally, one way to handle 3a-c in lpp might be that it is not conditional but depending on which fields are filled vs empty different logic is run. It means that some defaults would have to be None.

See more dicussions in https://github.com/diffpy/diffpy.utils/issues/45

yucongalicechen:
closes #143 
@sbillinge ready for review

Here's a gui snapshot:
gui

### DIFF
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -21,12 +21,13 @@ def _define_arguments():
                 "The filename(s) or folder(s) of the datafile(s) to load. "
                 "Required.\n"
                 "Supply a space-separated list of files or directories. "
+                "If a filename contains whitespace, enclose it in quotes. "
                 "Long lists can be supplied, one per line, "
                 "in a file with name file_list.txt. "
                 "If one or more directory is provided, all valid "
                 "data-files in that directory will be processed. "
                 "Examples of valid inputs are 'file.xy', 'data/file.xy', "
-                "'file.xy, data/file.xy', "
+                "'file.xy data/file.xy', "
                 "'.' (load everything in the current directory), "
                 "'data' (load everything in the folder ./data), "
                 "'data/file_list.txt' (load the list of files "
@@ -177,29 +178,27 @@ def _add_mud_selection_group(p, is_gui=False):
         **({"widget": "FileChooser"} if is_gui else {}),
     )
     g.add_argument(
-        "-td",
+        "-d",
         "--theoretical-from-density",
         help=(
             "Estimate mu*D theoretically using sample mass density. "
-            "Specify the sample composition (chemical formula), "
-            "incident x-ray energy in keV, "
-            "and sample mass density in g/cm^3 "
-            "in that exact order "
-            "and separated by commas with no whitespaces "
-            "(e.g., 'ZrO2,2,1.2')."
+            "Specify the chemical formula, incident x-ray energy (in keV), "
+            "and sample mass density (in g/cm^3), in that exact order, "
+            "separated by commas (e.g., ZrO2,20,1.5). "
+            "If you add whitespaces, "
+            "enclose it in quotes (e.g., 'ZrO2, 20, 1.5'). "
         ),
     )
     g.add_argument(
-        "-tp",
+        "-p",
         "--theoretical-from-packing",
         help=(
             "Estimate mu*D theoretically using packing fraction. "
-            "Specify the sample composition (chemical formula), "
-            "incident x-ray energy in keV, "
-            "and packing fraction (0 to 1) "
-            "in that exact order "
-            "and separated by commas with no whitespaces "
-            "(e.g., 'ZrO2,2,0.5')."
+            "Specify the chemical formula, incident x-ray energy (in keV), "
+            "and packing fraction (0 to 1), in that exact order, "
+            "separated by commas (e.g., ZrO2,20,0.5). "
+            "If you add whitespaces, "
+            "enclose it in quotes (e.g., 'ZrO2, 20, 0.5'). "
         ),
     )
     return p
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -152,7 +152,11 @@ def _define_arguments():
 def _add_mud_selection_group(p, is_gui=False):
     """Current Options:
     1. Manually enter muD (`--mud`).
-    2. Estimate muD from a z-scan file (`-z` or `--z-scan-file`).
+    2. Estimate from a z-scan file (`-z` or `--z-scan-file`).
+    3. Estimate theoretically based on sample mass density
+    (`-td` or `--theoretical-from-density`).
+    4. Estimate theoretically based on packing fraction
+    (`-tp` or `--theoretical-from-packing`).
     """
     g = p.add_argument_group("Options for setting mu*D value (Required)")
     g = g.add_mutually_exclusive_group(required=True)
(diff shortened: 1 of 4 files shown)


src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -165,10 +169,39 @@ def _add_mud_selection_group(p, is_gui=False):
...
+        help=(
+            "Estimate mu*D experimentally from a z-scan file. "
+            "Specify the path to the file "
+            "used to compute the mu*D value."
+        ),
         **({"widget": "FileChooser"} if is_gui else {}),
     )
+    g.add_argument(
+        "-td",
+        "--theoretical-from-density",
```
yucongalicechen, line 181: I thought about using four separate arguments for sample composition, energy, density, and packing fraction. But that doesn't work with ``add_mutually_exclusive_group``, which I like because it reinforces that only one of the four options can be selected, and it looks good on GUI too. So, I think combining three values into a single argument might actually be a better choice?

sbillinge: yes, I like this solution.  

Maybe to avoid user headaches you can allow whitespace and then strip it internally before parsing.

Also, I think that it will work without the quotes.  Could you test that?  As long as the parser is expecting a string, I think it will parse everything as a string until it encounters the next `--something` flag, but please check.  It is worth checking because it is a huge waste of time to have to always type those quotes

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -152,7 +152,11 @@ def _define_arguments():
 def _add_mud_selection_group(p, is_gui=False):
     """Current Options:
     1. Manually enter muD (`--mud`).
-    2. Estimate muD from a z-scan file (`-z` or `--z-scan-file`).
+    2. Estimate from a z-scan file (`-z` or `--z-scan-file`).
+    3. Estimate theoretically based on sample mass density
+    (`-td` or `--theoretical-from-density`).
```
sbillinge, line 157: check but this is breaking the usual pattern that it should be `-l` (one letter) or `--letters` (multiple letters need multiple dashes.

sbillinge (commented):
please see comments

src/diffpy/labpdfproc/tools.py
```diff
@@ -319,13 +319,13 @@ def _set_mud_from_zscan(args):
 
 def _parse_theoretical_input(input_str):
     """Helper function to parse and validate the input string."""
-    parts = input_str.split(",")
+    parts = [part.strip() for part in input_str.split(",")]
```
yucongalicechen, line 322: allow whitespace

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -177,29 +178,27 @@ def _add_mud_selection_group(p, is_gui=False):
...
+            "Specify the chemical formula, incident x-ray energy (in keV), "
+            "and sample mass density (in g/cm^3), in that exact order, "
+            "separated by commas (e.g., ZrO2,20,1.5). "
+            "If you add whitespaces, "
+            "enclose it in quotes (e.g., 'ZrO2, 20, 1.5'). "
         ),
     )
     g.add_argument(
-        "-tp",
+        "-p",
```
yucongalicechen, line 193: corrected

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -177,29 +178,27 @@ def _add_mud_selection_group(p, is_gui=False):
...
-            "incident x-ray energy in keV, "
-            "and sample mass density in g/cm^3 "
-            "in that exact order "
-            "and separated by commas with no whitespaces "
-            "(e.g., 'ZrO2,2,1.2')."
+            "Specify the chemical formula, incident x-ray energy (in keV), "
+            "and sample mass density (in g/cm^3), in that exact order, "
+            "separated by commas (e.g., ZrO2,20,1.5). "
+            "If you add whitespaces, "
+            "enclose it in quotes (e.g., 'ZrO2, 20, 1.5'). "
```
yucongalicechen, line 189: allow whitespace - edits in help message

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -21,12 +21,13 @@ def _define_arguments():
...
                 "Required.\n"
                 "Supply a space-separated list of files or directories. "
+                "If a filename contains whitespace, enclose it in quotes. "
                 "Long lists can be supplied, one per line, "
                 "in a file with name file_list.txt. "
                 "If one or more directory is provided, all valid "
                 "data-files in that directory will be processed. "
                 "Examples of valid inputs are 'file.xy', 'data/file.xy', "
-                "'file.xy, data/file.xy', "
+                "'file.xy data/file.xy', "
```
yucongalicechen, line 30: small error corrected - there should be no comma

yucongalicechen:
@sbillinge ready for another review! I addressed the comment about the whitespace and edited help messages accordingly.

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -21,12 +21,13 @@ def _define_arguments():
                 "The filename(s) or folder(s) of the datafile(s) to load. "
                 "Required.\n"
                 "Supply a space-separated list of files or directories. "
+                "If a filename contains whitespace, enclose it in quotes. "
```
sbillinge, line 24: maybe mention that, as a general matter, avoid filenames with whietspaces in, but if it does, then enclose in quotes

src/diffpy/labpdfproc/labpdfprocapp.py
sbillinge: What does the parser do if there is whitespace but no quotes.  Did you test that?

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -177,29 +178,27 @@ def _add_mud_selection_group(p, is_gui=False):
...
         "--theoretical-from-packing",
         help=(
             "Estimate mu*D theoretically using packing fraction. "
-            "Specify the sample composition (chemical formula), "
-            "incident x-ray energy in keV, "
-            "and packing fraction (0 to 1) "
-            "in that exact order "
-            "and separated by commas with no whitespaces "
-            "(e.g., 'ZrO2,2,0.5')."
+            "Specify the chemical formula, incident x-ray energy (in keV), "
```
sbillinge, line 197: since this seems to be used 2x, define it in a variable and call the variable.

sbillinge (commented):
please see comments

src/diffpy/labpdfproc/labpdfprocapp.py
yucongalicechen: @sbillinge ready for another review. I edited the help messages according to the comments above. When there's whitespace but no quotes the CLI will raise an error saying that it doesn't recognize the command

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -12,6 +12,13 @@
 from diffpy.utils.diffraction_objects import XQUANTITIES, DiffractionObject
 from diffpy.utils.parsers.loaddata import loadData
 
+theoretical_mud_hmsg_suffix = (
+    "in that exact order, "
+    "separated by commas (e.g., ZrO2,17.45,0.5). "
+    "If you add whitespaces, "
+    "enclose it in quotes (e.g., 'ZrO2, 17.45, 0.5'). "
+)
```
yucongalicechen, line 20: common part for help messages

PR #175, opened 2025-04-25
feat: initial commit on getmud module

### LINKED ISSUE #166: feat: getmud UC1
### Proposed solution

Implement this in a library format: return mu or D for given muD (we can reuse functions from computing theoretical muD).

yucongalicechen:
closes #166 
Initial commit. Probably needs a bit more discussion on what's the most useful here.

### DIFF
--- a/src/diffpy/labpdfproc/getmud.py
+++ b/src/diffpy/labpdfproc/getmud.py
@@ -1,0 +1,74 @@
+import numpy as np
+from scipy.optimize import newton
+
+from diffpy.utils.tools import compute_mu_using_xraydb
+
+
+def estimate_mass_density(mud, diameter, sample_composition, energy):
+    """Estimate sample mass density (g/cm^3) from mu*D
+    using capillary diameter, sample composition, and energy.
+
+    Parameters
+    ----------
+    mud : float
+        The given product of attenuation coefficient mu
+        in mm^{-1} and capillary diameter in mm.
+    diameter : float
+        The given diameter of the sample capillary in mm.
+    sample_composition : str
+        The chemical formula of the material.
+    energy : float
+        The energy of the incident x-rays in keV.
+
+    Returns
+    -------
+    estimated_density : float
+        The estimated mass density of the packed powder/sample in g/cm^3.
+    """
+    mu = mud / diameter
+
+    def residual_density(density):
+        return np.abs(
+            compute_mu_using_xraydb(
+                sample_composition, energy, sample_mass_density=density
+            )
+            - mu
+        )
+
+    estimated_density = newton(residual_density, x0=1.0, x1=10.0)
+    return estimated_density
+
+
+def estimate_diameter(
+    mud,
+    sample_composition,
+    energy,
+    sample_mass_density=None,
+    packing_fraction=None,
+):
+    """Estimate capillary diameter (mm) from mu*D and mu.
+
+    Parameters
+    ----------
+    mud : float
+        The given product of attenuation coefficient mu
+        in mm^{-1} and capillary diameter in mm.
+    sample_composition : str
+        The chemical formula of the material.
+    energy : float
+        The energy of the incident x-rays in keV.
+    sample_mass_density : float
+        The mass density of the packed powder/sample in g/cm^3.
+    packing_fraction : float, optional, Default is None
+        The fraction of sample in the capillary (between 0 and 1).
+        Specify either sample_mass_density or packing_fraction but not both.
+
+    Returns
+    -------
+    diameter : float
+        The given diameter of the sample capillary in mm.
+    """
+    mu = compute_mu_using_xraydb(
+        sample_composition, energy, sample_mass_density, packing_fraction
+    )
+    return mud / mu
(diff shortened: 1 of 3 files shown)


tests/test_getmud.py
```diff
@@ -0,0 +1,59 @@
...
+
+@pytest.mark.parametrize(
+    "inputs, expected_mass_density",
+    [
+        (
+            {
+                "mud": 2.0,
+                "diameter": 1.5,
+                "sample_composition": "ZrO2",
+                "energy": 20,
```
sbillinge, line 14: it is probably a better test if we use reasonable energies.  These would be Cu Kalpha, Mo Kalpha, Ag Kalpha.  They don't have to be exact but ballpark,

src/diffpy/labpdfproc/getmud.py
```diff
@@ -0,0 +1,74 @@
...
+
+    Returns
+    -------
+    diameter : float
+        The given diameter of the sample capillary in mm.
+    """
+    mu = compute_mu_using_xraydb(
+        sample_composition, energy, sample_mass_density, packing_fraction
+    )
+    return mud / mu
```
sbillinge, line 74: isn't mud/ mu = d?

src/diffpy/labpdfproc/getmud.py
```diff
@@ -0,0 +1,74 @@
...
+    packing_fraction : float, optional, Default is None
+        The fraction of sample in the capillary (between 0 and 1).
+        Specify either sample_mass_density or packing_fraction but not both.
+
+    Returns
+    -------
+    diameter : float
+        The given diameter of the sample capillary in mm.
+    """
+    mu = compute_mu_using_xraydb(
```
sbillinge, line 71: comment out and replace with pass.  Let's just work on tests to begin with, until we learn what we want the function to do.

src/diffpy/labpdfproc/getmud.py
```diff
@@ -0,0 +1,74 @@
...
+    packing_fraction=None,
+):
+    """Estimate capillary diameter (mm) from mu*D and mu.
+
+    Parameters
+    ----------
+    mud : float
+        The given product of attenuation coefficient mu
+        in mm^{-1} and capillary diameter in mm.
+    sample_composition : str
```
sbillinge, line 56: These are not needed presumably.  If I know mud and I know mu then nothing else matters.

sbillinge (commented):
please see comments.  I am a little confused what is going on here.  Do we have UCs for these functionalities?  What is the user wanting to do?  Once we know that we can design what the functions should do and capture that in tests.  Only then do we write any functions.

sbillinge:
> @sbillinge ready for review - I edited the docstrings and tests based on our UC1-2. For UC2 I kept it for now but I will send a message to @till-schertenleib to see if it makes sense. I moved UC3-4 to #181 for the next release since we do not have our density database now.

where can I find the UCs?  Please can you put a link here?

yucongalicechen:
> > @sbillinge ready for review - I edited the docstrings and tests based on our UC1-2. For UC2 I kept it for now but I will send a message to @till-schertenleib to see if it makes sense. I moved UC3-4 to #181 for the next release since we do not have our density database now.
> 
> where can I find the UCs? Please can you put a link here?

It's here: https://github.com/diffpy/diffpy.labpdfproc/issues/166.

sbillinge (commented):
OK, this is great now we have the UCs.  I think we are doing UC1 and UC3 and UC4.

UC1 is composition+energy+rho -> diameter
UC3 is comp + E -> mu (fully dense and 50% packing fraction)

I think that we probably want:
F1: comp + E + density -> mu
F2: comp + E -> estimated mu after looking up a CIF

Then we can satisfy 
UC1 as `muD/F1(comp,E,density)` 
UC4 as `muD/F2(comp,E)`
F2 satisfies UC3

Please could you close this and make one PR for each of those functions?

Do we also need a command line app that the user runs to get this info?  Let's put that on a separate PR, with the docs too.

It is already 3:20 so I guess we won't get this merged by 5.  Shall we push off that meeting until tomorrow?

yucongalicechen:
@sbillinge Just to clarify before I make any new PRs, I will do the following:

1. Make a PR for F1 which addresses UC1 => this is needed for tomorrow's release
2. Make a PR for F2 which addresses UC3 + 4 (#181)
3. Make an issue and PR for CLI app
4. (Done) Make an issue for docs (this is in #182)

sbillinge:
That looks about right, but I think there needs to be a CLI app for UC1 also, no? 

We should maybe discuss what the CLI should look like.  Are they separate apps that do those tasks, or do we want one entry point but give different behavior if different command line switches are used?  For example, for UC1 should the user type `get_diameter <composition> <energy> <mass-density>` or do we want them to type something like `labpdfproc --get-d <composition> <energy> <mass-density>` or sthg else?

yucongalicechen:
> That looks about right, but I think there needs to be a CLI app for UC1 also, no?
> 
> We should maybe discuss what the CLI should look like. Are they separate apps that do those tasks, or do we want one entry point but give different behavior if different command line switches are used? For example, for UC1 should the user type `get_diameter <composition> <energy> <mass-density>` or do we want them to type something like `labpdfproc --get-d <composition> <energy> <mass-density>` or sthg else?

Gotcha gotcha I think I'm on the same page now. I've created a PR for this (#184), so I'm closing this one.

Issue #193, opened 2025-06-08
feat: add ``getmud`` CLI for UC3 and UC4


yucongalicechen:
### Problem

UCs pasted from #181. Since the functions are already implemented, I think we can include the CLI structure in this release.

UC3: Compute muD from zscan
1. User has a zscan file, and wants to check muD.
2. User uploads zscan file to lpp.gm.
3. lpp.gm returns muD.

UC4: Compute muD theoretically
1. User specifies sample composition, energy, sample mass density, and diameter.
2. User provides this information to lpp.gm.
3. lpp.gm returns muD.

Issue #192, opened 2025-06-08
docs: update documentation to reflect changes in the CLI structure and mud computation


yucongalicechen:
### Proposed solution

Ensure all revelant texts are updated to address changes made in issues #187, #188, and #191.

Issue #181, opened 2025-05-02
feat: getmuD UC3


yucongalicechen:
### Problem

UC3 (From #166):
1. User specifies sample composition and energy.
2. User provides this information labpdfproc.getmud (lpp.gm)
3. User receives (multiple) recommendations for mu depending on phase

### Proposed solution

lpp defaults packing fraction to 0.5 and 1 and computes using the function in diffpyutils), so the code can be something like:
```
def estimate_mu(sample_composition, energy):
    mu1 = compute_mu_using_xraydb(sample_composition, energy, packing_fraction=0.5)
    mu2 = compute_mu_using_xraydb(sample_composition, energy, packing_fraction=1)
    return mu1, mu2
```

sbillinge:
looks good.   I slightly modified UC3 based on our conversation.  We have to decide an algorithm for computing mu when there are more than one phases returned by the database.  There are different ways of doing it, we can discuss.

yucongalicechen:
@sbillinge Could you check the UCs I updated in the top comment based on our conversation today? Thanks. I have closed a few other issues as they were a bit outdated and confusing.

sbillinge:
One final thought.  If it were me I would be a bit lazy and not want to run `getmud` to get the muD, then have to write it down or remember it and then have to run `labpdfproc` again to get my corrected data, so we may want to offer this to them:
```
labpdfproc correct --data-file nacl-100K.xy -s NaCl -e Mo -d 1
```
and have it compute mud and run the correction.  It can assume a packing fraction of 0.5 but allow this to be overwritten?

cadenmyers13:
@sbillinge same thing here as here: https://github.com/diffpy/diffpy.labpdfproc/issues/193#issuecomment-3975074684

sbillinge:
closing as implemented

PR #196, opened 2025-07-04
feat: add ``applymud`` subcommand

### LINKED ISSUE #191: feat: new CLI architecture
> > I can merge this if you want (it is conflicted so that needs to be fixed) but my concern is that the CLI is not the way we want it so we can't release this way. Do you want to fix things on this PR or another one?
> 
> I can resolve the conflicts here so we can merge this PR, and address the CLI issue in a separate PR. Do you prefer to keep the different muD computation methods within getmuD only (as we discussed in #181)? I think it is also helpful to give users the option to compute muD during the correction.

yes, I am working off the understanding that we are being guided by the discussion in #181.  However, if we have this structure for `getmud` it kind of implies that there will be a similar construction for something with a name like `getcorr` or `applymud`, unless you can think of a better way of doing it.
`labprdfproc getmud -c NaCl -e 25`
`labpdfproc applymud [file1.xy, file2.xy] --mud 1.8`
`labprdfproc applymud [file1.xy, file2.xy] -c NaCl -e 25 -d 1.5`
something like that (this is not a complete list but just examples).

_Originally posted by @sbillinge in https://github.com/diffpy/diffpy.labpdfproc/issues/189#issuecomment-2886458149_

yucongalicechen:
closes #191

The `getmud` subcommand is tracked through #193 and #181. Docs update for this PR is tracked through #192.

### DIFF
--- a/src/diffpy/labpdfproc/labpdfprocapp.py
+++ b/src/diffpy/labpdfproc/labpdfprocapp.py
@@ -73,6 +73,7 @@ def _define_arguments():
                 "If the specified directory doesn't exist it will be created."
             ),
             "default": None,
+            "widget": "DirChooser",
         },
         {
             "name": ["-x", "--xtype"],
@@ -158,7 +159,7 @@ def _define_arguments():
     return args
 
 
-def _add_mud_selection_group(p, is_gui=False):
+def _add_mud_selection_group(p, use_gui=False):
     """Current Options:
     1. Manually enter muD (`--mud`).
     2. Estimate from a z-scan file (`-z` or `--z-scan-file`).
@@ -173,7 +174,7 @@ def _add_mud_selection_group(p, is_gui=False):
         "--mud",
         type=float,
         help="Enter the mu*D value manually.",
-        **({"widget": "DecimalField"} if is_gui else {}),
+        **({"widget": "DecimalField"} if use_gui else {}),
     )
     g.add_argument(
         "-z",
@@ -183,7 +184,7 @@ def _add_mud_selection_group(p, is_gui=False):
             "Specify the path to the file "
             "used to compute the mu*D value."
         ),
-        **({"widget": "FileChooser"} if is_gui else {}),
+        **({"widget": "FileChooser"} if use_gui else {}),
     )
     g.add_argument(
         "-d",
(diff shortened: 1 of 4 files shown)


src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -158,7 +159,7 @@ def _define_arguments():
     return args
 
 
-def _add_mud_selection_group(p, is_gui=False):
+def _add_mud_selection_group(p, use_gui=False):
```
yucongalicechen, line 162: Changed ``is_gui`` to ``use_gui`` because I feel like it's a better variable name

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -207,39 +208,50 @@ def _add_mud_selection_group(p, is_gui=False):
     return p
 
 
-def get_args(override_cli_inputs=None):
-    p = ArgumentParser()
-    p = _add_mud_selection_group(p, is_gui=False)
+def _register_applymud_subparser(subp, use_gui=False):
```
yucongalicechen, line 211: Here it's basically the same as before, but I'm taking it out as a separate function for adding arguments to ``applymud``. I think it'll be cleaner when define another one for ``getmud`` later.

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -207,39 +208,50 @@ def _add_mud_selection_group(p, is_gui=False):
...
 
 
-@Gooey(required_cols=1, optional_cols=2, program_name="labpdfproc GUI")
-def gooey_parser():
-    p = GooeyParser()
-    p = _add_mud_selection_group(p, is_gui=True)
-    for arg in _define_arguments():
-        kwargs = {key: value for key, value in arg.items() if key != "name"}
-        p.add_argument(*arg["name"], **kwargs)
+def create_parser(use_gui=False):
```
yucongalicechen, line 224: Function for creating gooey/argument parser

src/diffpy/labpdfproc/tools.py
```diff
@@ -45,6 +45,7 @@
     "wavelength",
     "theoretical_from_density",
     "theoretical_from_packing",
+    "subcommand",
```
yucongalicechen, line 48: Do not write subcommand in the output files

sbillinge:
This looks good.  I recommend we work a bit on the argparse helps so they are more readable when they are in the gui.

Do we have docs?  Some of the longer explanations can go to the docs if we have them.  Or is there a way to have them longerin in the cli but shorter in the gui?  That could work too.

For the top input,  in the gui it is not so helpful a name for the variable.  maybe `data_sets` or something is more descriptive.

yucongalicechen:
> This looks good. I recommend we work a bit on the argparse helps so they are more readable when they are in the gui.
> 
> Do we have docs? Some of the longer explanations can go to the docs if we have them. Or is there a way to have them longerin in the cli but shorter in the gui? That could work too.
> 
> For the top input, in the gui it is not so helpful a name for the variable. maybe `data_sets` or something is more descriptive.

Sounds great! I opened a new issue for this #197

PR #199, opened 2025-07-15
skpkg: apply black to all files in the project directory

Frankiecesco510:


sbillinge:
this seems to contain edits that were not done by black.  I can only merge black changes if there are no human edits.  

Let's close this and do things one step at a time.   In the PR  Please post a screenshot of the command you ran and a bit of the output you get from git diff.  Then post a screenshot of what you get from `git status` and we can navigate our way through this.

PR #206, opened 2026-01-21
build: Add support for python 3.14 and remove support for python 3.11

### LINKED ISSUE #203: Add support for python 3.14 and remove support of python 3.11
To add and remove support of these python versions, open `pyproject.toml` and first replace this line,
```
requires-python = ">=3.11, <3.14"
```
with this line

```
requires-python = ">=3.12, <3.15"
```
Then, remove this line

```
        'Programming Language :: Python :: 3.11',
```
and add this line

```
        'Programming Language :: Python :: 3.14',
```

Finally, do a global search in the repo to make sure nothing else needs to be changed.

stevenhua0320:
@sbillinge Ready to review, Closes #203 
P.S After the workflow run, I realized that `diffpy.labpdfproc` still uses the old version of workflow for `test-on-pr`, that is why the python version is hard-coded in the workflow and did not run the latest version of python.

### DIFF
--- a/news/py314.rst
+++ b/news/py314.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Support for Python 3.14
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* Support for Python 3.11
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ maintainers = [
 description = "Tools for processing x-ray powder diffraction data from laboratory sources."
 keywords = ['powder xrd', 'absorption correction', 'pdf', 'diffpy']
 readme = "README.rst"
-requires-python = ">=3.11, <3.14"
+requires-python = ">=3.12, <3.15"
 classifiers = [
         'Development Status :: 5 - Production/Stable',
         'Environment :: Console',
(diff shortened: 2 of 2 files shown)


cadenmyers13:
@stevenhua0320 yes that would be great if you have the bandwidth. Let me know if not

stevenhua0320:
@sbillinge I think this one could be repackaged using `package update` right? I could do that to make this package up to date with data structure. All I need to do is to open new PRs and make as small edits to content as possible.

sbillinge:
> @sbillinge I think this one could be repackaged using `package update` right? I could do that to make this package up to date with data structure. All I need to do is to open new PRs and make as small edits to content as possible.

that is correct.  There will be very little to be done I think, so it may all go in on a single PR, but just see.  After this is done we can be sure htat everything is correctly aligned.  However, before you do that, I think there was some work we were doing on scikit-package.  Please could you look over there and see if there are any open PRs or issues?  It would be good to close them and then use this to test any changes didn't break anything.

sbillinge:
I think this can be fixed by pip installing gooey.  Let's look into this.

sbillinge:
I think I would like to do the full release of 0.3.0 at 3.11-3.13 and then we can merge this and immediately do a 0.4.0 release that is 3.12 - 3.14.  This has the advantage that there is a version that covers 3.11 to 3.14, but also that the current version is following the group standard so we don't have to keep track of some packages that diverge.

sbillinge:
@stevenhua0320 I think this will be superseded by the work being done by Adib but please see if anything here can be cherry picked or whatever.  Let me know if you want this closed (or merged)

PR #212, opened 2026-02-15
skpkg: run `package update` for scikit-package 0.3.0

cadenmyers13:
This also acts as a test for scikit-package 0.3.0 conda-forge release as mentioned here https://github.com/scikit-package/scikit-package/issues/668#issuecomment-3904549408

### DIFF
--- a/.github/workflows/tests-on-pr.yml
+++ b/.github/workflows/tests-on-pr.yml
@@ -1,15 +1,53 @@
 name: Tests on PR
 
 on:
+  push:
+    branches:
+      - main
   pull_request:
   workflow_dispatch:
 
 jobs:
-  tests-on-pr:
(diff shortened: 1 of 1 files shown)


cadenmyers13:
@sbillinge ready for review. `package update` is awesome now! with the new `cookiecutter.json` this command is super handy and pretty quick as long as you stay on top of the skpkg changes

cadenmyers13:
> so I think we may want to put in a shift and work on getting a release of Gooey out the door.

@sbillinge yeah that would be good. Can we do a minimal release of it to get it up to latest Python? We could support 3.11-3.14 since there is no 3.11 releas?

cadenmyers13:
@sbillinge okay I reverted tests-on-pr back and made an issue to update it when `gooey` is released

PR #219, opened 2026-03-03
docs: Fix API Docs rendering 

cadenmyers13:


cadenmyers13:
@sbillinge Before rc release check this please. 

I was having problems with API docs rendering and mocking utils seemed to fix it. Im not too sure exactly if we need to do that or have it in docs.txt. Do you know? 

edit: removed it from docs.txt, dont think thats right

PR #223, opened 2026-04-01
feat: Check if saved files exists and raise an early error if they do

### LINKED ISSUE #220: docs: some little documentation tweaks for the 0.4.0 release
### Problem
no problem, just a few cleaning suggestions.

### Proposed solution
 - remove the note that the gui requires Python <=3.12.  This is not true if the installation instructions install gooey using pip
 - Maybe have a getting started section that just points to the example.  Just easier to navigate.
 - Let's change the filenames of the example files to have dashes instead of underscores (new group standard).  Also, let's change that the corrected file is called `<name>-corrected.chi` and not `<name>_corrected.chi` (this would be a small code change).  Also, I dislike this name because it doesn't capture what was corrected.  Let's think about it but maybe `<name>-abs-corrected-<mur-value>.chi` 
 - The section that begins `To save the correction file, specify the -c or --output-correction flag,...` there is no text about why one might want to do that, and that in general it is not needed.
 - The reviewers of the paper wanted us to move everywhere to MuR from MuD.  I don't love that, but it may be ambiguous if it is sometimes mur and sometimes mud.  Shall we change everywhere to mur?
 - z-scan is introduced everywhere as if everyone knows what that is, but noone does.  It is well described in the paper so we don't really have to reproduce that here.  But a sentence about what it is, and a mention to read the paper, when it is first introduced both at the top of the instructions and also in the example section would be good.  A link to the diffpy.utils docs is present (good) but also mention the paper.
 - instead of `<name>_cve.chi` for the correction file, again, how about making it less cryptic, something like `<name>-cve-correction-<mur-value>.chi`.   We could call it "abs-correction" but then the difference between the corrected data and the correction itself would be hard to differentiate because the difference would just be `abs-corrected` vs `abs-correction`.  Maybe calling it `mur-correction` instead of `cve-correction` though?

cadenmyers13:
closes #220 

Before, this was just a print statement which was confusing. Now an error is raised if the output file already exists.


cadenmyers13:
@sbillinge ready for review

cadenmyers13:
Ah, it's failing

src/diffpy/labpdfproc/labpdfprocapp.py
```diff
@@ -131,20 +131,14 @@ def _add_credit_args(parser, use_gui=False):
 
 
 def _save_corrected(corrected, input_path, args):
-    outfile = args.output_directory / (input_path.stem + "_corrected.chi")
-    if outfile.exists() and not args.force:
-        print(f"WARNING: {outfile} exists. Use --force to overwrite.")
-        return
+    outfile = args.output_directory / (input_path.stem + "-mud-corrected.chi")
```
sbillinge: no biggie, but as a general rule, let's compose as f-strings, not using +

src/diffpy/labpdfproc/tools.py
```diff
@@ -528,6 +528,27 @@ def load_package_info(args):
     return args
 
 
+def _check_saved_file_exists(args):
+    """Check if the output files already exist based on the input paths
+    and output directory."""
+    existing_files = []
+    for path in args.input_paths:
+        outfile = args.output_directory / (path.stem + "-mud-corrected.chi")
```
sbillinge: see above

sbillinge (commented):
I reviewed the ocde. modulo a couple of small comments it looks good.  Let me know when tests are passing and I can merge.

